PR #45, opened 2014-12-11
Updated typescript definitions inline with r3

lukephills:


lukephills:
No worries and yes, please do.

PR #51, opened 2015-03-07
Make compatible with CommonJS and Browserify

adamrenklint:
Ran into some trouble when trying to use this library in a Browserify build, so here's a patch. Hope you accept the changes. Cheers!

- export the main Tone object on ```module```, if it exists
- fix incorrect ```main``` path in ```package.json```

### DIFF
--- a/build/Tone.js
+++ b/build/Tone.js
@@ -14336,6 +14336,8 @@
 		define( "Tone", [], function() {
 			return Tone;
 		});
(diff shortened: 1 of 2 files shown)


tambien:
thanks for the fix! I've added the CommonJS line of code to `grunt/fragments/after.frag` and it will appear in the next build. 

Please make future PRs on the dev branch.

PR #60, opened 2015-03-27
Fix phase bug for complex waveforms

mindofmatthew:
Fix bug where phase is applied to partials relative to the partial,
rather than the fundamental, frequency


tambien:
Thanks for your contribution, but please make the PR to the `dev` branch.

PR #61, opened 2015-03-27
Phase patch

mindofmatthew:
Fix bug where phase is applied to partials relative to the partial,
rather than the fundamental, frequency


tambien:
Thanks Matthew! great catch

PR #87, opened 2015-08-19
Tone p5

polyrhythmatic:
I changed the gulpfile to create both p5.Tone and Tone in both minified and unminified versions. This could also probably be done cleaner or as a separate command so let me know your thoughts.


tambien:
Thanks for the PR! Please take out all of the build files (Tone.Preset, Tone.js, Tone.min and p5.Tone.js) and only include the p5.frag and the changes to the gulpfile. Also please make the PR to the `dev` branch instead of master.

PR #88, opened 2015-08-19
P5 tone

polyrhythmatic:
removed build files

commit to dev


polyrhythmatic:
tested with polyrhythmatic/p5.js/866 👍

PR #89, opened 2015-08-20
updated Microphone, added ExternalInput

polyrhythmatic:
Here it is! Let me know if you think the external input example is too pedantic. I wanted to demonstrate how you can enumerate device ID's and set them. You can also stop and start the audio stream and change the inputNum.

### DIFF
--- a/Tone/source/ExternalInput.js
+++ b/Tone/source/ExternalInput.js
@@ -1,0 +1,220 @@
+define(["Tone/core/Tone", "Tone/source/Source"], function(Tone){
+
+	"use strict";
+
+	/**
+	 *  @class  Tone.ExternalInput is a WebRTC Audio Input. Check 
+	 *          [Media Stream API Support](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream_API)
+	 *          to see which browsers are supported. As of
+	 *          writing this, Chrome, Firefox, and Opera 
+	 *          support Media Stream. Chrome allows enumeration 
+	 *          of the sources, and access to device name over a 
+	 *          secure (HTTPS) connection. See [https://simpl.info](https://simpl.info/getusermedia/sources/index.html) 
+	 *          vs [http://simple.info](https://simpl.info/getusermedia/sources/index.html) 
+	 *          on a Chrome browser for the difference.
+	 *         
+	 *  @constructor
+	 *  @extends {Tone.Source}
+	 *  @param {number} [inputNum=0] If multiple inputs are present, select the input number. Chrome only.
+	 *  @example
+	 *  var motu = new Tone.ExternalInput(3);
+	 *  var soundflower = new Tone.ExternalInput().connect()
+	 *  
+	 *  Tone.ExternalInput.onload = function(){
+	 *  	motu.start();
+	 *  	selectSoundflower(Tone.ExternalInput.sourceList);
+	 *  }
+	 *  	
+	 *  function selectSoundflower(sources){
+	 *  	for(var i = 0; i < sources.length; i++){
+	 *  		if(sources[i].label === "soundflower"){
+	 *  			soundflowerIn.inputNum = i;
+	 *  			soundflowerIn.start();
+	 *  		}
+	 *  	}
+	 *  };
+	 */
+
+	Tone.ExternalInput = function(){
+
+		var options = this.optionsObject(arguments, ["inputNum", "onload"], Tone.ExternalInput.defaults);
+		Tone.Source.call(this, options);
+
+		/**
+		 *  @type {MediaStreamAudioSourceNode}
+		 *  @private
+		 */
+		this._mediaStream = null;
+		
+		/**
+		 *  @type {LocalMediaStream}
+		 *  @private
+		 */
+		this._stream = null;
+		
+		/**
+		 *  @type {Object}
+		 *  @private
+		 */
+		this._constraints = {"audio" : true};
+
+		if(Tone.ExternalInput.sourceList.length === 0){
+			Tone.ExternalInput._getSources();
+		} 
+
+		/**
+		 * The input source position in Tone.ExternalInput.sourceList. 
+		 * Set before ExternalInput.start().
+		 * @type {[type]}
+		 */
+		this.inputNum = options.inputNum;
+
+	};
+
+	Tone.extend(Tone.ExternalInput, Tone.Source);
+
+	/**
+	 * the default parameters
+	 * @type {Object}
+	 */
+	Tone.ExternalInput.defaults = {
+		"inputNum" : 0,
+		"onload" : Tone.noOp
+	};
+
+	/**
+	 * Start the stream
+	 * @private
+	 */
+	Tone.ExternalInput.prototype._start = function(){
+		if(!Tone.ExternalInput.sourceList.length && Tone.ExternalInput._hasGetSources && this.inputNum){
+			console.error("Default source selected, use ExternalInput.onload to allow source loading");
+		}
+		if(!Tone.ExternalInput._hasGetSources && this.inputNum > 0){
+			console.error("This browser does not support MediaStreamTrack sourceId");
+		}
+		if (this.inputNum < Tone.ExternalInput.sourceList.length && this.inputNum > -1) {
+			this._constraints = {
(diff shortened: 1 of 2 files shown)


polyrhythmatic:
I'm just now seeing a note about an example. Do you want a full-fledged example for the example page?

Tone/source/ExternalInput.js
```diff
@@ -0,0 +1,220 @@
...
+	 *  			soundflowerIn.inputNum = i;
+	 *  			soundflowerIn.start();
+	 *  		}
+	 *  	}
+	 *  };
+	 */
+
+	Tone.ExternalInput = function(){
+
+		var options = this.optionsObject(arguments, ["inputNum", "onload"], Tone.ExternalInput.defaults);
```
tambien: You never do anything with the "onload" parameter. Probably should be taken out.

Tone/source/ExternalInput.js
```diff
@@ -0,0 +1,220 @@
...
+	 * @type {Array}
+	 */
+	Tone.ExternalInput.sourceList = [];
+
+	/**
+	 * indicates whether browser supports MediaStreamTrack.getSources (i.e. Chrome vs Firefox)
+	 * @type {Boolean}
+	 * @private
+	 */
+	Tone.ExternalInput._hasGetSources = true;
```
tambien: You could check if it has the `getSources` method here. instead of once you call `_getSources` instead of setting it to true initially. 

```
Tone.ExternalInput._hasGetSources = Tone.prototype.isFunction(MediaStreamTrack.getSources)
```

also not really into the name `hasGetSources`, maybe `canGetSources`?

Tone/source/ExternalInput.js
```diff
@@ -0,0 +1,220 @@
...
+		"onload" : Tone.noOp
+	};
+
+	/**
+	 * Start the stream
+	 * @private
+	 */
+	Tone.ExternalInput.prototype._start = function(){
+		if(!Tone.ExternalInput.sourceList.length && Tone.ExternalInput._hasGetSources && this.inputNum){
+			console.error("Default source selected, use ExternalInput.onload to allow source loading");
```
tambien: `throw new Error("error message here")` instead of just reporting the error to the console.

polyrhythmatic: A side effect of this is that Tone will no longer select the microphone as default for an input, which you want it to do in the case of firefox and opera since they have their own source selection interface. But given how the different browsers handle external input differently, it is hard to standardize this feature.

polyrhythmatic: Now I'm thinking it may be best to remove the errors. If the user is on Mozilla they can select the input they want through the browser interface (this._constraints.audio.optional.sourceId does nothing) and if they're on Chrome, they can use Tone to select the inputNum.

tambien:
ok. now i've got a big wrench to throw in this whole thing: what if instead of the static onload callback and all the stuff that goes into that, you had to explicitly `open` and `close` the stream. and the `open` method could take a callback as an argument. It could populate the sourcesList (if that hasn't been done already). Then the `start` and `stop` methods would simply be muting and unmuting the microphone output. 

```
Tone.ExternalInput.prototype.open = function(callback){
  if (Tone.ExternalInput.sourceList.length === 0){
    Tone.ExternalInput.getSources(function(){
      navigator.getUserMedia(this._constraints, function(stream){
        this._onStream(stream);
        callback();
      }.bind(this), function(err){
        throw new Error(err);
      });
    }
  }
}

Tone.ExternalInput.prototype.close = function(){
  if (this._stream){
    this._stream.stop();
  }
}

Tone.ExternalInput.prototype._start = function(time){
  this._gate.gain.setValueAtTime(1, time);
}

Tone.ExternalInput.prototype._stop = function(time){
  this._gate.gain.setValueAtTime(0, time);
}
```

tambien:
decoupling the `open`/`close` from the `start`/`stop` makes the microphone more consistent with the other sources (especially the ability to schedule one or more starts and stops in the future). 

the current implementation would break if you tried to do this because the stream would open and immediately close (or close before it's even open and probably throw an error that `this._stream` is `null`):
```
mic.start().stop("+2")
```

but my proposed changes would make this sort of thing possible.

polyrhythmatic:
I tried implementing this, but I think the problem is that the callback is for the getSources function, so we need to account for this too. we can't set the _constraints objects until after the getsources function has been called.

I think I can figure this out - I'll work on it tomorrow

tambien:
doesn't nesting the callbacks like i wrote above account for this?

polyrhythmatic:
Yea I guess you're right, I'm still not able to set the right inputNum. The correct constraints object is being passed in but I can't figure out why it isn't being set. It's using the default input number regardless.

polyrhythmatic:
Figured it out! I was showing 6 sources, because when I was populating the sourceList i was not sorting for type. Added this line that I misplaced along the way...

```
MediaStreamTrack.getSources(function (media_sources){
				for(var i = 0, max = media_sources.length; i < max; i++) {
					if(media_sources[i].kind === "audio"){
					Tone.ExternalInput.sourceList[i] = media_sources[i];
					}
				}
				callback();
			});
```

I'll submit another PR with my changes soon!

polyrhythmatic:
Since we are eliminating the onLoad function, its now not possible to look at the sourceList values and choose an input. I think this is an important function like I was talking about - this allows the user to mimic the firefox/opera functionality.

We could make the _getSources function public, let me know your thoughts on this. I think it would be pretty nice to be able to use the callback to populate a dropdown or automatically select "soundflower" or something like that.

polyrhythmatic:
@tambien I've updated ExternalInput, would love your thoughts on the getsources function being public.

Tone/source/ExternalInput.js
```diff
@@ -0,0 +1,235 @@
...
+	 *  @return {Tone.ExternalInput} this
+	 */
+	Tone.ExternalInput.prototype.open = function(callback){
+		if(Tone.ExternalInput.sourceList.length === 0){
+			Tone.ExternalInput.getSources(function(){
+				this._getUserMedia(callback);
+			}.bind(this));
+		} else {
+			this._getUserMedia(callback);
+		}
```
tambien: should `return this`

Tone/source/ExternalInput.js
```diff
@@ -0,0 +1,235 @@
...
+	 *  @param  {function =} callback The callback function to 
+	 *                       execute when the stream is open
+	 *  @return {Tone.ExternalInput} this
+	 */
+	Tone.ExternalInput.prototype.open = function(callback){
+		if(Tone.ExternalInput.sourceList.length === 0){
+			Tone.ExternalInput.getSources(function(){
+				this._getUserMedia(callback);
+			}.bind(this));
+		} else {
```
tambien: could be simplified:

Tone.ExternalInput.getSources(function(){
  this._getUserMedia(callback);
}.bind(this));

Tone/source/ExternalInput.js
```diff
@@ -0,0 +1,235 @@
...
+		this._stream = null;
+		
+		/**
+		 *  @type {Object}
+		 *  @private
+		 */
+		this._constraints = {"audio" : true};
+
+		if(Tone.ExternalInput.sourceList.length === 0){
+			Tone.ExternalInput.getSources(function(){});
```
tambien: This isn't necessary. It should be explicitly invoked by the user.

Tone/source/ExternalInput.js
polyrhythmatic: added

Tone/source/ExternalInput.js
polyrhythmatic: removed

Tone/source/ExternalInput.js
polyrhythmatic: change added 👍

polyrhythmatic:
Applied these changes, and changed sourceList to sources like we discussed

I would still like to add the ability to set input by name, id, or input number but I feel like this may take me longer than I have time for right now.

Also, do we want to take into consideration that webrtc will soon be secure connection only? Do you have any more information on this? I can't find any.

If this is the case, we may end up changing our method a little bit.

PR #92, opened 2015-08-25
Cymbal synth

polyrhythmatic:
Just a cursory commit so I can get your thoughts. I anticipate having to do a bit more work on this.

I switched the oscillators to FMSynths and it sounds really interesting. The PWM oscillators sound great as well and I am open to having a constructor function to choose one or the other, if you think they need more available variation.

I couldn't come up with another name for harmonicity that wasn't vague like "metallic" or "timbre" or "character". These sort of describe the sound it provides though.

I've also been toying around with the idea of making a handle for the harmonicity of the FMSynths. These can influence the sound nicely.

You'll notice on line 142 I do something a bit hacky to start the oscillators

```
			this._oscillators[i].triggerAttack(Tone.context.currentTime);
```

Couldn't figure out a way around it and I'm not sure if it could be problematic.

Here's a link of what I was basing this from

http://media.soundonsound.com/sos/jul02/images/fig14.gif

For the dispose method - I can't get all of the methods inside of CymbalComponent to dispose of themselves. Is this a problem?

Lastly - should I document the components inside the CymbalComponent class? Not sure about how you would do this.

Seth


tambien:
This is looking good. 

I will add my comments in the code. Would you please remove all other changed files but the CymbalSynth?

I think the trick with the oscillators is just not to use an FMSynth which includes a bunch of extra stuff like an OmniOscillator and multiple envelopes. You can create a frequency modulated oscillator just using two oscillators connected to a Tone.Multiply. 

i think we should go with `inharmonicity` instead of `harmonicity` since everywhere else harmonicity is used to mean a frequency multiple and here it's a 0-1 range. 

The dispose should definitely clean up everything by disconnecting or disposing them. I think the reason that you can't get everything to dispose is because you should put [this](https://github.com/polyrhythmatic/Tone.js/blob/cymbal-synth/Tone/instrument/CymbalSynth.js#L298) line above [this one](https://github.com/polyrhythmatic/Tone.js/blob/cymbal-synth/Tone/instrument/CymbalSynth.js#L276). otherwise the dispose method that you just defined is overwritten.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+		 *  @private
+		 *  @type {Array}
+		 */
+		this._freqMult  = [];
+
+		//make all the FMSynths, connect the frequency signals
+		for(var i = 0; i < 6; i++){
+			this._scaledSignals[i] = new Tone.Scale(this.inharmRatios[i], this.harmRatios[i]);
+			this._freqMult[i] = new Tone.Multiply();
+			this._oscillators[i] = new Tone.SimpleFM({
```
tambien: Make your own frequency modulated oscillator instead of using the FMSynth.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+		}).connect(this.output);
+		this.envelope = new Tone.AmplitudeEnvelope({
+			"attack" : options.envelope.attack,
+			"decay" : options.envelope.decay,
+			"sustain" : 0,
+			"release": 0,
+			"attackCurve" : "exponential"
+
+		}).connect(this._highPass);
+		this.input = new Tone.Filter({
```
tambien: i would name this something more reasonable like `filter` and then just make the input point to that.

`this.filter = this.input = new Tone.Filter({...`

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+			"type" : "bandpass",
+			"Q" : options.Q
+		}).connect(this.envelope);
+		this.cutoff = new Tone.Signal(options.cutoff).connect(this.input.frequency);
```
tambien: It's unnecessary to make another signal here. You could just have the cutoff point to the this.input.frequency

`this.cutoff = this.input.frequency`.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+		this._highPassFrequency = new Tone.Multiply(options.frequencyScalar).connect(this._highPass.frequency);
+		this.cutoff.connect(this._highPassFrequency);
+		this.resonance = new Tone.Signal(options.resonance).connect(this.input.Q);
```
tambien: See the above note about the `cutoff`

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+		this._highPassFrequency.dispose();
+		this._highPassFrequency = null;
+		this.resonance.dispose();
+		this.resonance = null;
+		this._highPassResonance.dispose();
+		this._highPassResonance = null;
+		return this;
+	};
+
+	Tone.extend(CymbalComponent);
```
tambien: This line needs to be above `CymbalComponent.prototype.dispose`, otherwise that method will just be overwritten.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+	 * @private
+	 */
+	var CymbalComponent = function(options){
+		this.output = this.context.createGain();
+		this.volume = new Tone.Signal({
+			"param" : this.output.gain, 
+			"units" : Tone.Type.Decibels,
+			"value" : options.volume
+		});
+		this._readOnly(["volume"]);
```
tambien: Not really necessary to make this readOnly since it's a totally internal interface and you don't need to worry about misusing it internally.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+			}
+		}).connect(this.output);
+
+		/**
+		 *  [harmonicity description]
+		 *  @type {Positive}
+		 *  @signal
+		 */
+		this.harmonicity = new Tone.Signal(options.harmonicity);
+		this.harmonicity.units = Tone.Type.Positive;
```
tambien: you could assign the units in the constructor: 

`new Tone.Signal(options.harmonicity, Tone.Type.Positive)`

I think the type is actually `NormalRange` since it is connected to a `Tone.Scale` which only takes values between 0-1.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+				"attack": options.strike.envelope.attack, 
+				"decay": options.strike.envelope.decay, 
+			}
+		}).connect(this.output);
+
+		/**
+		 * The body portion of the cymbal hit.
+		 * @type {CymbalComponent}
+		 */
+		this.body = new CymbalComponent({
```
tambien: You could just pass in the entire options object and let the CymbalComponent manage all this stuff. Maybe leave out the `resonanceScalar` and `frequencyScalar` and pass those in as the next arguments.

`new CymbalComponent(options.body, 1.643, 0.333);`

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+			"cutoff" : 7000,
+			"resonance" : 6,
+			"envelope" : {
+				"attack" : 0.01,
+				"decay" : 1.5
+			}
+
+
+		},
+	};
```
tambien: I like these defaults! they're simple, understandable and expressive. 👍

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+
+	/**
+	 *  Trigger the release of the cymbal.
+	 *  Can be used to [choke](https://en.wikipedia.org/wiki/Cymbal_choke) 
+	 *  the cymbal
+	 *  @param  {Time} [time=now] the time the cymbal will release
+	 *  @returns {Tone.CymbalSynth} this
+	 *  @example
+	 *  //set the release to a small amount for a realistic choke
+	 *  cymbal.body.envelope.release = 0.2;
```
tambien: I might switch the default envelope release to something like 0.2 instead of 0. That way `triggerRelease` is useful right away without having to change any settings. Should probably put envelope release in the defaults then as well.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+		return this;
+	};
+
+	/**
+	 * Cymbal part helper class
+	 * @private
+	 */
+	var CymbalComponent = function(options){
+		this.output = this.context.createGain();
+		this.volume = new Tone.Signal({
```
tambien: This might be a good place to use `Tone.Volume`

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+			"volume" : options.strike.volume,
+			"cutoff" : options.strike.cutoff,
+			"frequencyScalar" : 1.886,
+			"resonance" : options.strike.resonance,
+			"resonanceScalar" : 0.533,
+			"envelope" : {
+				"attack": options.strike.envelope.attack, 
+				"decay": options.strike.envelope.decay, 
+			}
+		}).connect(this.output);
```
tambien: See the note below. This could be written just `new CymbalComponent(options.strike, 1.886, 0.533);`

Tone/instrument/CymbalSynth.js
polyrhythmatic: change made, added default for body release 0.3

Tone/instrument/CymbalSynth.js
polyrhythmatic: makes sense, but shouldn't it be private? like this._filter?

Tone/instrument/CymbalSynth.js
polyrhythmatic: removed

Tone/instrument/CymbalSynth.js
polyrhythmatic: technically it's normal range, but it sounds cool beyond 0-1. I should probably just add that to the documents.

Tone/instrument/CymbalSynth.js
polyrhythmatic: added

Tone/instrument/CymbalSynth.js
polyrhythmatic: change added

Tone/instrument/CymbalSynth.js
tambien: it could be. but given that the whole thing is a private class. it doesn't really matter. just be consistent.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+			"attack" : options.envelope.attack,
+			"decay" : options.envelope.decay,
+			"sustain" : 0,
+			"release": 0,
+			"attackCurve" : "exponential"
+
+		}).connect(this._highPass);
+		this.input = new Tone.Filter({
+			"type" : "bandpass",
+			"Q" : options.Q
```
tambien: I don't think you pass in an `options.Q` do you?

Tone/instrument/CymbalSynth.js
polyrhythmatic: added

Tone/instrument/CymbalSynth.js
polyrhythmatic: added

Tone/instrument/CymbalSynth.js
polyrhythmatic: added this change, and I realized I can't dispose of the output gain node I added. Right now I'm just setting it equal to null

tambien: If it's a native node, disconnect it and then set it to null. 

You can also call `Tone.prototype.dispose.call(this)` which will do that for you on the input and output nodes.

Tone/instrument/CymbalSynth.js
polyrhythmatic: I was setting it in the filter constructor

```
		this.input = new Tone.Filter({
			"type" : "bandpass",
			"Q" : options.resonance
		}).connect(this.envelope);
```

but now I'm setting it directly on the signal
```
		this.resonance = this.input.Q;
		this.resonance.value = options.resonance;
```

Tone/instrument/CymbalSynth.js
polyrhythmatic: the reason I made this private is because there is no need to access any of its attributes directly. the highpass filter frequency and q are both modified via scalars on the main frequency and resonance values. so i don't really want people changing these since there is not much to change (perhaps the rolloff but thats it)

tambien: Oh right. i forgot that these are actually exposed as `body` and `strike`. Yes everything that is not accessed like `envelope`, `resonance`, etc should be private and `_underscorePrefixed`.

Tone/instrument/CymbalSynth.js
polyrhythmatic: great, added the Tone.prototype.dispose line and works like a charm 👍

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+	Tone.CymbalSynth.prototype.dispose = function(){
+		Tone.Instrument.prototype.dispose.call(this);
+		this.harmonicity.dispose();
+		this.harmonicity = null;
+		this.frequency.dispose();
+		this.frequency = null;
+		this.inharmRatios = null;
+		this.harmRatios = null;
+
+		for(var i = 0; i <6; i++){
```
tambien: nitpicky, but probably better as `this._oscillators.length` instead of `i < 6`.

Tone/instrument/CymbalSynth.js
```diff
@@ -0,0 +1,304 @@
...
+		this.frequency = new Tone.Signal(options.frequency, Tone.Type.Frequency);
+
+		/**
+		 *  Inharmonic ratio of frequencies based on the Roland TR-808
+		 *  Taken from https://ccrma.stanford.edu/papers/tr-808-cymbal-physically-informed-circuit-bendable-digital-model
+		 *  @private
+		 *  @static
+		 *  @type {Array}
+		 */
+		this.inharmRatios = [1.0, 1.483, 1.932, 2.546, 2.630, 3.897];
```
tambien: since the `inharmRatio` and `harmRatios` are actually private, static, constant, they should not be put in the constructor function. I would pull them out before the constructor and set them with `var`. Take a look at the way it's done in [Tone.Freeverb](https://github.com/Tonejs/Tone.js/blob/c94861ca5495185206bfb3b03eb957bd44610d00/Tone/effect/Freeverb.js#L21)

Tone/instrument/CymbalSynth.js
polyrhythmatic: great, I updated the names to be a bit more concise

Tone/instrument/CymbalSynth.js
polyrhythmatic: change added

Tone/instrument/CymbalSynth.js
polyrhythmatic: Do you mean actually connect two oscillators into one signal

```
this.multiply = new Tone.Multiply().connect(this.strike);

this.osc1 = new Tone.Oscillator().connect(this.multiply, 0, 0);
this.osc2 = new Tone.Oscillator().connect(this.multiply, 0, 1);
```

I've tried a few configurations but I may just end up copying the SimpleFM to maintain the sound I have now but cut back on the unnecessary envelopes. 

With just two oscillators and a Tone.Multiply, how do you control ModulationIndex and Harmonicity?

polyrhythmatic: I'm still running into some issues with this not sounding the same, and weird things like signal values such as _freqMult[i] being zero when they should be the frequency value for each oscillator

```		/**
		 *  The level of inharmonicity between the frequencies of the oscillators. 
		 *  Interesting sounds can be had between 0 and 20.
		 *  @type {Positive}
		 *  @signal
		 */
		this.inharmonicity = new Tone.Signal(options.inharmonicity, Tone.Type.Positive);

		/**
		 *  The base frequency for the series of oscillators. 
		 *  All oscillators scale in a harmonic or inharmonic pattern
		 *  depending on the value of the inharmonicity.
		 *  @type {Frequency}
		 *  @signal
		 */
		this.frequency = new Tone.Signal(options.frequency, Tone.Type.Frequency);

		/**
		 *  A series of FMSynths
		 *  @private
		 *  @type {Array}
		 */
		this._carriers = [];

		this._modulators = [];

		/**
		 *  Tone.ScaledSignals used for interpolating between harmonic
		 *  and inharmonic ratios
		 *  @private
		 *  @type {Array}
		 */
		this._scaledSignals = [];

		/**
		 *  Tone.Multipliers for frequency multiplication of scaled signals
		 *  @private
		 *  @type {Array}
		 */
		this._freqMult  = [];

		this.harmonicity = [];

		this._modulationNodes = [];

		this._modulationIndex = [];

		//make all the FMSynths, connect the frequency signals
		for(var i = 0; i < 6; i++){
			this._scaledSignals[i] = new Tone.Scale(inharmRatios[i], harmRatios[i]);
			this._freqMult[i] = new Tone.Multiply();
			this._carriers[i] = new Tone.Oscillator();
			this._carriers[i].type = "square";
			this._carriers[i].volume.value = -10;
			
			this._modulators[i] = new Tone.OmniOscillator();
			this._modulators[i].type = "pulse";
			this._modulators[i].width.value = 0.75;
			this._modulators[i].volume.value = -10;

			this.harmonicity[i] = new Tone.Multiply(1.5);

			this._modulationIndex[i] = new Tone.Multiply(10);


			this.inharmonicity.connect(this._scaledSignals[i]);
			this._scaledSignals[i].connect(this._freqMult[i], 0, 0);
			this.frequency.connect(this._freqMult[i], 0, 1);

			this._modulationNodes[i] = this.context.createGain();
			this._modulators[i].connect(this._modulationNodes[i].gain);

			this._freqMult[i].chain(this.harmonicity[i], this._modulationNodes[i]);
			this._freqMult[i].chain(this._modulationIndex[i], this._modulationNodes[i]);
			this._modulationNodes[i].gain.value = 0;

			this._modulationNodes[i].connect(this._carriers[i].frequency);
			this._freqMult[i].connect(this._carriers[i].frequency);
			this._carriers[i].connect(this.strike);
			this._carriers[i].connect(this.body);

			this._carriers[i].start();
			this._modulators[i].start();
		}
	};

```

Here's what I have so far. I'm really stuck on debugging this. Any idea what I'm doing wrong?

tambien:
Thanks so much for your implementation. I ended up adapting a lot of pieces of it and putting it into [Tone.MetalSynth](https://github.com/Tonejs/Tone.js/blob/dev/Tone/instrument/MetalSynth.js) after creating an underlying FMOscillator.

PR #99, opened 2015-10-21
Changed defaultArg implementation and fixed Tone.Buffer init with AudioBuffer

lickylick:
I changed the defaultArg implementation, because it was causing issues with instanceof checks. Because the defaultArg created a copy of the given object, but stripping of any object type (like AudioBuffer) it would cause checks further down the line to fail. For instance in Tone.Buffer implementation there is a check if the given object is an AudioBuffer but this check would always fail because defaultArg would strip the AudioBuffer type of the object. I think the defaultArg implementation should do what it says, only set a value on a property of the given object if that property is not yet set (i.e. a default value). It is simple and should not break any other code in the project.

I also fixed a bug in Tone.Buffer where this._buffer.set is called where it should be this.set.

### DIFF
--- a/Tone/core/Buffer.js
+++ b/Tone/core/Buffer.js
@@ -64,7 +64,7 @@ define(["Tone/core/Tone"], function(Tone){
 		this.onload = options.onload.bind(this, this);
 
 		if (options.url instanceof AudioBuffer){
-			this._buffer.set(options.url);
+			this.set(options.url);
 			this.onload(this);
 		} else if (typeof options.url === "string"){
 			this.url = options.url;
--- a/Tone/core/Tone.js
+++ b/Tone/core/Tone.js
@@ -489,13 +489,7 @@ define(function(){
 	///////////////////////////////////////////////////////////////////////////
 
 	/**
-	 *  if a the given is undefined, use the fallback. 
-	 *  if both given and fallback are objects, given
-	 *  will be augmented with whatever properties it's
-	 *  missing which are in fallback
-	 *
-	 *  warning: if object is self referential, it will go into an an 
-	 *  infinite recursive loop. 
+	 *  Add default values for the properties that are not yet set
 	 *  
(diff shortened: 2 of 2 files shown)


tambien:
Thanks for the PR and for trying to address this issue. `defaultArg` is a little more complicated because it will also handle nested objects. This is useful when passing in complicated presets (like Tone.DuoSynth for example) which can have a lot of nested parameters. But i agree this is an issue when you don't want it to create a copy if a class instance is passed in. 

I've tried addressing this issue by checking if the `given` and `fallback` are [object literals and not class instances](https://github.com/Tonejs/Tone.js/blob/dev/Tone/core/Tone.js#L509). This seems to work in my [test](https://github.com/Tonejs/Tone.js/blob/dev/test/core/Buffer.js#L60). 

And thanks for the Tone.Buffer fix, but I had already fixed it. :)

4b78dcbed10259ab4258fef5afd0b529ca6793e0

lickylick:
Thanks for the fast response, I didn't know the full scope of how the defaultArg method works, your solution seems solid. 

One thing to point out in dev is that optionsObject calls defaultArg(options, defaults, shallowCopy) while the shallowCopy arg is not present in the defaultArg declaration. But probably you are still working on this. 

Do you have any idea when you are going to merge these changes into master?

tambien:
Good catch! just removed that. 

I'll merge dev into master when i make a new release. I'm planning on releasing r6 sometime in the Fall. I've still got more work to do on the "musical structures" classes that i've added plus a few small improvements/updates to the core.

lickylick:
Looking forward to the new release, also want to say I really appreciate the work you put in Tone, I'm just diving into web audio and javascript (coming from native sound synthesis with c, c++), but Tone is easy to use and the source code is well structured and easy to understand

tambien:
That's very nice of you to say. Thank you.

PR #103, opened 2015-12-06
Tremolo stereo spread control

lukephills:
Hi Yotam,

I really like the new stereo Tremolo but I found that I wanted to get the original mono sound from the r5. I've added a spread option which allows the user to control the panning of the two LFO's

At 0 both LFO's are centered. At 180 both LFOs are panned hard left and right respectively. 

What do you think?


tambien:
This is awesome! so simple and flexible. thanks!

PR #115, opened 2016-01-31
fixed start offset in Tone.Part, now behaves like Tone.Player

### LINKED ISSUE #113: Allow Tone.Part.start() offset to be before loopStart
It looks like the offset in `Tone.Part.start()` only works if it's after loopStart. Would it be possible to implement the same kind of behaviour as `Tone.Player`? 

For example, a looped `Tone.Part` with a length of 2m and loopStart at 1m, would only loop the last bar. With an offset of 0, Tone.Part would play from the beginning, but will start looping after the first bar.

Thank you!

hiddedejong:
Hi Yotam,

I tried to fix #113 myself and it seems to work on my machine! I hope this is useful for you.

Hidde

### DIFF
--- a/Tone/event/Part.js
+++ b/Tone/event/Part.js
@@ -146,7 +146,11 @@ define(["Tone/core/Tone", "Tone/event/Event", "Tone/core/Type", "Tone/core/Trans
 	Tone.Part.prototype.start = function(time, offset){
 		var ticks = this.toTicks(time);
 		if (this._state.getStateAtTime(ticks) !== Tone.State.Started){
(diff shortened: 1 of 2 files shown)


tambien:
Thanks Hidde. This looks good. Would you run the tests in the events folder and let me know if they all pass? 

cd into the gulp folder and type `gulp collectTests --event` then run the `index.html` file in the test folder.

hiddedejong:
All tests passed. I also added a new test which is now also included in the pull request!

tambien:
thanks Hidde!

PR #127, opened 2016-02-26
adds FMbell

polyrhythmatic:
Something I was working on last year and forgot about. This may be able to be refactored so it uses less components, but I really like how it ended up sounding. Try adding a curve to the modulation index from 10 to 1 over the life of the note, or playing with the harmonicity on a curve. I think it has lots of potential for many different sounds! 

The one thing I struggle with is this is a percussive sound but it's pitched, so right now "note" length is dictated by the decay value on the envelopes and the sustain is set to 0. Curious about what you think about this.


tambien:
Used this as reference for both the MetalSynth and the FMOscillator so thanks. In retrospect, i should have merged your changes and then altered them so that you would have gotten due github credit. Sorry about that.

PR #128, opened 2016-03-03
tried adding an example to pattern

evejweinberg:
example from class


tambien:
thanks for the PR. A few comments:

Please make all Pull Requests on the `dev` branch; this one is on `master`. 

Also the example uses Tone.Pattern, but the class file is for Tone.CtrlPattern. Could you update the example to use Tone.CtrlPattern instead of Tone.Pattern?

Also each new line has to start with a `* ` with exactly one space for unindented lines and a space + tab for indented lines otherwise the formatting will be off when it's compiled into the docs. I use DocBlockr and it does that formatting automatically for me.

PR #141, opened 2016-05-16
Close previous audio context

frankbaele:
Close previous audio context before assigning a new one:

api: https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/close

Ran into this issue when multiple embed codes of my application where used on 1 page.
Example usecase: http://stackoverflow.com/questions/25977269/using-audiocontext-in-multiple-iframes

### DIFF
--- a/Tone/core/Tone.js
+++ b/Tone/core/Tone.js
@@ -765,6 +765,10 @@ define(function(){
 	 *  @param {AudioContext} ctx The new audio context to set
 	 */
 	Tone.setContext = function(ctx){
+		//First close the previouse audiocontext.
+		if(typeof Tone.context.close !== 'undefined'){
+			Tone.context.close();
+		}
 		//set the prototypes
 		Tone.prototype.context = ctx;
 		Tone.context = ctx;


tambien:
Thanks for the PR. You bring up a good point about creating too many contexts in this situation, but i feel like since `setContext` does not create a context, it is more the responsibility of the part of the application that opened the context to close it. 

Though, maybe since Tone creates a context by default, it _is_ it's responsibility to close it before setting a new one. Maybe a way of making it more explicit by adding an optional second argument  `Tone.setContext(ctx, closePrevious)`. 

But i'm not sure that closing the previous context when setting a new one would fix the issue in the SO question. it seems like you hit the limit of the number of AudioContexts on one page and the not because you were trying to set the context too many times without closing the previous one (i might be mistaken). i think the solution would be to remove a previous iframe to make sure there is no more than 6. and potentially close the AudioContext when doing so (but my guess is that this would already be done by just removing the iframe). 

A compatibility issue with this change is that many of the tests currently set an OfflineContext, run a test and the re-set the previous context... though this could be rewritten i suppose...

Also as a side note, please make future PRs to the dev branch.

frankbaele:
ok, maybe it was a bad pull request and the iframe problem is totally irrelevant.

But it did bring up the underlying problem that tone.js is creating an audio context by default. 
Which is good for out of the box usage. But if you set a new one you are ending up with 2 audio contexts. What is just a waste since there is a hard limit on audiocontexts.

Another solution would be that you are allowed to set an audiocontext before tone initialises, like most web audio plugins do.

Frank

PR #149, opened 2016-07-16
Updated node versions

amilajack:


tambien:
Closing out some old PRs. Thanks for the contribution, but since the code relies very little on the node.js version, i don't see a reason to test on both Node 4 and 6. Probably better just to pick the latest and stick to that.

PR #157, opened 2016-08-11
Source.sync() design

taylorbf:
Changes design of Source.sync(), so that a synced player will play at a certain timeline position, rather than in relation to the transport’s start time.

API is still:

player.sync(“1m”)

Added Transport.seek() method and updated the looping mechanism to use .seek(). The .seek() method is created in order to start/stop synced players. This could also be done in Transport.position’s setter, but I ran into some stack overflows.

The loop timing is not perfect, to my ear; you may be able to find some improvements.

I didn’t run the build script yet; I will leave that to you if you incorporate these edits.

### DIFF
--- a/Tone/core/Transport.js
+++ b/Tone/core/Transport.js
@@ -206,11 +206,10 @@ function(Tone){
 			tickTime += Tone.Time(this._swingTicks * 2/3, "i").eval() * amount;
 		} 
 		//do the loop test
-		if (this.loop){
-			if (ticks === this._loopEnd){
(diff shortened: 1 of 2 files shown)


tambien:
Thanks for the PR. I revisiting Source syncing to the transport. It is now much more precise and powerful, capable of updating to the latest position of the Transport and updating to the correct offset in the player.

here's an example

https://tonejs.github.io/examples/daw.html

ChenPaulYu:
Can you give the codepen or codesandbox example ?

PR #162, opened 2016-09-03
Minor fix (proposal)

kant:


tambien:
It seemed like one of those Markdown links was still broken even with this PR since I made a big parenthesis mess initially. Thanks for bringing my attention to it, i think it's fixed now.

PR #176, opened 2016-10-31
Change docs to correct default swingSubdivision

taekyunggg:
The default swingSubdivison for the Transport is an 8th note, not a 16th note.


tambien:
thanks!

PR #220, opened 2017-03-24
BoundLFO

k23Burke:
I wanted to get this PR up so we can start discussing more specifics of the code itself. I'm currently using this build in one of my projects so I know that it is functional but there might be some modifications we can make going forward as well as examples/testing to include before this is ready to be pulled in. Nonetheless this will give us something a bit more concrete to work on.

1. I tried utilizing the Add/Multiply/Scale components both before and after the WaveShaper however I only got the output I was looking for when I used the WaveShaper by itself.

2. I first attempted to have all the perameters (min, max, lowBound, highBound) set individually - however getters and setters didn't make sense in this instance largely because the WaveShaper would have to be re-calculated on any parameter change. This means individually setting the four params would lead to some heavier processing. Maybe I should provide getters and setters in case you wish to change a param individually and suggest that users set the param object at once. Curious as to your thoughts.

If you want to see the BoundLFO in action, shoot me a dm and I can get you a link.


tambien:
Thanks for the PR @k23Burke. I'm not going to merge this one, but it would be great if you could turn some of this work into a module like Tone.Constrain or Tone.Clip, i think it would be clearer for most people to use and more general purpose. I feel like a BoundLFO is a fairly specific use-case and that the API of having a min/max as well as a min/maxBound is a little confusing. I think it'd be clearer to pass an LFO through something like `Tone.Clip(50, 2000)` which will constrain the output to that range instead of having those attributes be in the LFO itself. 

Thanks again for your contribution.

k23Burke:
@tambien I tried circling around to this one again but after trying to hack away with the Web Audio API docs and the Tone codebase, I'm not making much progress here. I'd love to contribute but I haven't found a solid solution for this one.

Doing a value lookup in order to constrain the param would be too costly in processing and without knowing the minimum and maximum values, potentially outside of the bounds, it would be impossible to constrain the values properly with a waveshaper.

Curious as to your strategy on this.

PR #222, opened 2017-03-26
AmplitudeEnvelope: Fix wrong default value in doc

timpulver:
`new Tone.AmplitudeEnvelope().get("sustain")` returns `0.5`, not `1.0`


tambien:
Thanks @timpulver, but this example was meant to illustrate how to use the AmplitudeEnvelope, it isn't meant to show the default values. To see the defaults for this class you have to click on [Envelope](http://tonejs.github.io/docs/#Envelope) which is a little confusing. I am eventually planning to refactor the docs code to account for these sorts of things and i'll try to make it more clear in the future.

PR #257, opened 2017-08-24
MultiPlayer fixes

k23Burke:
I was messing around with multiplayer enough to at least contribute this. See comment in the thread concerning choking with Tone.Players


tambien:
Thanks for the PR, but Multiplayer has been deprecated in r11, so i am going to close this.

MangoSmoothie101010:
How to make sure n (n>1) sounds are started at the same time using `PLayers`?

tambien:
@MangoSmoothie101010 easiest way is to simply pass in the same time into the `start` method for all of the players.

PR #262, opened 2017-09-26
Fix typo in MembraneSynth description

seutje:
"applied during during ..." => "applied during ..."

Apologies if this is not the source to edit.


tambien:
thanks!

PR #297, opened 2018-01-31
Add some infos about how to use the toolchain

tchakabam:
Hey there :) 

As not using gulp in any of my projects before, and generally information lacking on the setup here, I found it might be nice to leave some info to whoever might want to step in.

A few things noticed (or questions):

* I stated in the commit that running `gulp` as a system install will not run your local gulp install. But maybe I am wrong and gulp is smart about detecting a local install? I would need to correct that statement then.

* The root package `build` script does the above, or tries to rely on npm to resolve it to the local bin-path. Not sure what the intention is here?

* At least my npm (5.6.0) will not be smart about that for me, so basically when gulp is not installed system-wide, i get: 

```
> cd gulp && gulp

sh: 1: gulp: not found
```

That is probably because npm only resolves script shell to the local bin-path of the package script from where it was called.

* Therefore, it would probably be more "stable" to change your build and test scripts to something like: `cd gulp && ./node_modules/.bin/gulp`, or introduce a short-hand script in the root-package that calls local gulp bin explicitely, like `npm run gulp`, and to avoid making a system-wide gulp install necessary. 

You could then run individual tasks like this `npm run gulp -- lint` for example?

2. Don't commit build files

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,34 @@ Tone.js is a Web Audio framework for creating interactive music in the browser. 
 
 [Full Installation Instruction](https://github.com/Tonejs/Tone.js/wiki/Installation).
 
+# Dev
+
+If you want to contribute to Tone or create your own build you need to install the dev-dependencies.
+
+We are using [gulp](https://github.com/gulpjs/gulp), and keep the whole toolchain in it's own place.
+
+To install the development toolchain go from project root:
+
+```
+cd gulp
+npm install
+```
+
+Now for running the default pipeline with the local toolchain:
+
+```
+./node_modules/.bin/gulp
+```
+
+This will produce a minified distribution bundle in the `build` folder.
+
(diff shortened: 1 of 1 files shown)


tambien:
Thanks, but i think the better place to contribute this information would be on [this wiki page](https://github.com/Tonejs/Tone.js/wiki/Building). Perhaps a link should be added to the bottom of the README in the contributing section, since it's not currently linked to from the README directly.

tchakabam:
@tambien Ok sure :) What about the questions around `gulp` that arised for me ?

tambien:
@tchakabam yeah i've been meaning to do a bit of a refactor on the toolchain part of things. You're totally right that it only works if gulp is installed globally. It'd be more convenient to modify it as you say and to include more of the scripts in the root level package.json.

PR #318, opened 2018-03-14
Encode URL when loading buffer

### LINKED ISSUE #317: Buffer loading doesn't handle special characters
Not sure how common an issue this is, and can usually be worked around just by renaming files, but thought I'd bring it up as I just ran into it.

Here's an example of the bug, with the url passed in being `mp3/snare #1.mp3`:
http://usdivad.com/misc/tone_buffer_example.html

And an example with the issue fixed (all the code on the page is the same, just swapping out the Tone.js library):
http://usdivad.com/misc/tone_buffer_example_fixed.html

^ So in the second link, `http://usdivad.com/misc/mp3/snare #1.mp3` gets converted to `http://usdivad.com/misc/mp3/snare%20%231.mp3`

usdivad:
Fixes #317.

A consideration would be whether there are any special characters, other than `/`, that we *don't* want `encodeURIComponent()` to handle?

### DIFF
--- a/Tone/core/Buffer.js
+++ b/Tone/core/Buffer.js
@@ -466,7 +466,18 @@ define(["Tone/core/Tone", "Tone/core/Emitter", "Tone/type/Type", "Tone/shim/Audi
 		}
 
 		var request = new XMLHttpRequest();
-		request.open("GET", encodeURIComponent(Tone.Buffer.baseUrl + url), true);
+
+		// encode the url, separating by slash
+        var urlSplit = (Tone.Buffer.baseUrl + url).split("/");
+        var encodedUrl = "";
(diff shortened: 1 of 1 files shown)


tambien:
Thanks!

tambien:
This causes an error when a complete url path is passed in: i.e. `https://example.com/audio.mp3` -> `https%3A//example.com/audio.mp3`.

usdivad:
Ah rats, so sorry about that! I've added a quick fix in https://github.com/usdivad/Tone.js/commit/8b2c5f3ea4cf3e140501fc3104f7fe3e56ab9fde, but no worries if you think this whole URL encoding matter creates more issues than it solves.

PR #329, opened 2018-04-20
Fix AudioNode.disconnect error when disconnecting from signals/params

jbpuryear:
WebAudio's AudioNode.disconnect will throw an error if specifying an input number when disconnecting from an AudioParam (at least in chrome and firefox for linux). This pull request checks if we're disconnecting from an AudioParam before calling WebAudio's native disconnect method, and calls it without specifying an input if we are.


tambien:
thanks!

PR #335, opened 2018-04-25
Sampler Instrument can play entire sample, even when pitch shifting down. 

### LINKED ISSUE #334: Sampler instrument truncates playback when pitch shifting down
The Sampler instrument will truncate the playback of a sample when pitch shifting down. This happens with the `TriggerAttack` and `TriggerAttackRelease` methods -- (you cannot play the full sample even by setting a long `duration`)

What do you think about doing something like this?

https://github.com/CharlesHolbrow/Tone.js/commit/ebc255d8b1f54e875346b9addd455343d8ee8b22

Most commercial samplers I've used (like Kontakt) default to this behavior - or provide an option to pitch shift without time-stretching.

CharlesHolbrow:
fixes #334


tambien:
thank you!

CharlesHolbrow:
Wow, that was quick. Thanks! looking forward to seeing this on NPM!

PR #336, opened 2018-04-26
fix Tone.Midi.toFrequency example

joeLepper:
Hi there.

I noticed that the `Tone.Midi.toFrequency` example in the docs seems to be copypasta from `Tone.Midi.toMidi`.


joeLepper:
Closing in favor of #337.

Issue #345, opened 2018-05-12
feature request/discuss: polyphonic api for samplers


zfan40:
I noticed there is a handy way to trigger polyphonic synth with feeding notes array.
`poly.triggerAttack(["Ab3", "C4", "F5"], undefined, 0.2);`

Well, there is no such syntactic sugar for samplers. Do you think supporting that meaningful?
If so, I might be able to raise a PR.

tambien:
good idea. PR please!

PR #346, opened 2018-05-13
support sampler polyphonic syntax

zfan40:
feature mentioned in https://github.com/Tonejs/Tone.js/issues/345


tambien:
thanks!

christianhengst:
Hey! This is cool! Would be cool if I could use the

`polySampler.releaseAll()`

function as well. Currently it is not working, or am I missing something?

Issue #384, opened 2018-09-04
Memory usage for Noise


paulvarache:
Looking at the Memory usage of a page loading Tone.js (Only loading, not creating anything), I can see around 5MB of ram is use to store the noise arrays (pink, brown, white).

The source of this memory usage is located here: https://github.com/Tonejs/Tone.js/blob/502727736000dcd367db3492fc43f9ad850bf362/Tone/source/Noise.js#L188

Moving this execution to happen on demand (Creating the sample needed when used) would free some memory in project not using Noise at all, or using only one type of noise.

What would be the performance impact of this? Would it slow down the first usage of Tone.Noise? And if so, is this processing happening on every page using Tone.js just after it has been downloaded?

![image](https://user-images.githubusercontent.com/5550089/45029962-3ea30000-b042-11e8-8995-b6dbe785a7c0.png)

tambien:
On my computer it takes < 10ms to generate all three noise buffers, so approx 3ms per buffer. i think my concern when i set it up this way was that it could cause a performance related pop when it was first generated. this probably isn't an issue most of the time and i guess the hit for every user is probably greater than than a small startup hiccup when a user first initializes Tone.Noise

PRs are greatly appreciated if this is a priority for your application. i think the best way to tackle it would be to include those buffers as a private static property only when the first Tone.Noise is constructed.

paulvarache:
Thanks for your reply, this is great. I was worried deferring this could cause more issues. I will look into making a PR for this

PR #386, opened 2018-09-07
Make noiseArrays generate samples lazily

paulvarache:
Make `_noiseArrays`'s pink, brown and white properties getters with cached values. This leaves the signature of `_noiseArrays` identical but creates the arrays only when needed.
Related to #384

### DIFF
--- a/Tone/source/Noise.js
+++ b/Tone/source/Noise.js
@@ -185,56 +185,74 @@ define(["../core/Tone", "../source/Source", "../core/Buffer",
 	 *  @private
 	 *  @type {Array}
 	 */
-	var _noiseArrays = {
-		"pink" : (function(){
-			var buffer = [];
-			for (var channelNum = 0; channelNum < channels; channelNum++){
-				var channel = new Float32Array(bufferLength);
-				buffer[channelNum] = channel;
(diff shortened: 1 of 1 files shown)


tambien:
thanks for these changes! will debug that safari failure, seems unrelated to your changes

paulvarache:
Thanks for the update. It seems I can run the test, but get an error while trying to start the headless chrome. It is worth mentioning that I'm running this on Windows 10. As for building, I still get a bunch of errors after following the instructions: `Module not found: Error: Can't resolve './typeType.js' in 'C:\Users\Paul\Documents\workspace\Tone.js\Tone'`.
I will open another Issue if I can't figure out what's failing

tambien:
oh good to know. it is probably a file path thing which i've only dealt with in a mac/linux environment. could probably be fixed by using node.js' `path` object instead of trying to spell out any paths. Any research you could do with that would be helpful. i don't have easy access to a windows machine

timothyboie:
https://github.com/Tonejs/Tone.js/blob/34e26059cabc0a42c58e43066c2a8205b94f1bec/scripts/collect_deps.js#L18

Add this line after:
reqString = reqString.replace(/\\/g, "/");

PR #387, opened 2018-09-10
Normalize paths for builds on windows

paulvarache:
Transforms backslashes to forward slashes to fix builds on windows


AdenFlorian:
@tambien 
@paulvarache 's solution works for me (windows 10, node v10.14.1), but your commit 8070135 doesn't. There are a couple open issues on path.posix for windows:
- https://github.com/nodejs/node/issues/13887
- https://github.com/nodejs/node/issues/13683

When using posix, for me, it's outputting things like this into `index.js` when building:
`require("./../../C:/Users/Me/Tone.js/Tone/component/Follower");`

Think we could go with paul's solution for now?

PR #394, opened 2018-10-03
Bump `webpack-cli` version to 3.1.2

Ameobea:
* This fixes an incompatibility that broke the build process that was introduced in webpack version 4.20.0: https://github.com/webpack/webpack/issues/8082


tambien:
thanks!

PR #410, opened 2018-12-12
Remove subtle blip on waveform in analysis example

ericandrewlewis:
There's a subtle blip at the beginning of the waveform on the analysis example page due to a special case for the first point on the drawn line:

screen shot 2018-12-11 at 10 10 17 pm

This PR fixes that blip:

screen shot 2018-12-11 at 10 12 31 pm


ericandrewlewis:
This is my first time using this library and it is super duper - thanks so much for building it!

ericandrewlewis:
It looks like the Travis CI build failed due to timeouts unrelated to this change.

tambien:
thank you!

PR #421, opened 2019-01-07
WIP: Replace patches

chrisguttandin:
Hi @tambien, I hope you're doing good.

this is not meant to be a mergeable pull request. It is more like a proof of concept which could be extended to other parts of Tone.js if you like it.

I recently noticed that Tone.js doesn't work well with my [standardized-audio-context](https://github.com/chrisguttandin/standardized-audio-context) package. After some debugging I figured that Tone.js is modifiying built-in objects like the AudioContext prototype. My library on the other hand uses these objects to determine the API support in the currently used browser.

Therefore I began to replace every modification of built-in objects with internal helper functions. For now I only replaced the chain() and connect() functions of AudioNodes.

I would argue that every modification of a built-in object should ideally be replaced with a helper function as it is not considered a good practice to modify objects owned by the browser. Tone.js is so popular that it might one day cause a similar problem as MooTools did by modifying the Array.prototype. (https://esdiscuss.org/topic/having-a-non-enumerable-array-prototype-contains-may-not-be-web-compatible)

Do you although think that using internal helpers is a viable approach? I'm curious on your thoughts.

### DIFF
--- a/Tone/component/AmplitudeEnvelope.js
+++ b/Tone/component/AmplitudeEnvelope.js
@@ -40,7 +40,7 @@ define(["../core/Tone", "../component/Envelope", "../core/Gain"], function(Tone)
 		 */
 		this.input = this.output = new Tone.Gain();
 
-		this._sig.connect(this.output.gain);
+		Tone.connect(this._sig, this.output.gain);
 	};
 
 	Tone.extend(Tone.AmplitudeEnvelope, Tone.Envelope);
--- a/Tone/component/CrossFade.js
+++ b/Tone/component/CrossFade.js
@@ -82,10 +82,10 @@ define(["../core/Tone", "../signal/Signal", "../signal/Subtract",
 		//connections
 		this.a.connect(this.output);
 		this.b.connect(this.output);
-		this.fade.chain(this._equalPowerB, this.b.gain);
+		Tone.chain(this.fade, this._equalPowerB, this.b.gain);
 		this._one.connect(this._invert, 0, 0);
 		this.fade.connect(this._invert, 0, 1);
-		this._invert.chain(this._equalPowerA, this.a.gain);
+		Tone.chain(this._invert, this._equalPowerA, this.a.gain);
 		this._readOnly("fade");
 	};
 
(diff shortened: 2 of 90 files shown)


Idicious:
Is far as I know Tonejs doesn't alter the prototype of native webaudio nodes. All Tone classes are wrapper classes around the native nodes, which internally call the native functions on them.

chrisguttandin:
Hi @Idicious, thanks for taking a look at this.

You're right replacing .chain() was a bit to eager. It only ends up on the prototype of the internal AudioNode. But .connect() and .disconnect() for example do get patched on the built-in prototype.

If you navigate for example to https://tonejs.github.io/examples/#oscillator in Chrome. You can open the dev tools and log `AudioNode.prototype.connect` it will not show the native function anymore.

This is the part of Tone.js which patches the AudioNode prototype: https://github.com/Tonejs/Tone.js/blob/dev/Tone/core/Context.js#L594

Idicious:
@chrisguttandin I see, I agree that not changing the native prototypes is better. Let's see what @tambien thinks of it.

tambien:
@chrisguttandin thanks for this PR. i totally agree i'd like to move away from changing any native nodes. i think it'd be pretty easy to replace the few instances of `chain`, `connect`, and `fan` that operate on native nodes because (as @Idicious mentioned) almost everything is wrapped. So i don't think that every instance of those methods needs to be replaced with a `Tone.connect`, but instead change the ones that operate on native nodes to something like `node.connect(toneNode.input)`. would require far fewer changes and does not introduce any new APIs. 

How am i only now seeing standardized-audio-context?! this is an amazing effort and i'd love to move Tone.js to rely on this instead of the very incomplete shims that i've got in there now. i've been looking through it this morning, i will follow up on your github with a couple more questions about integrating your work.

chrisguttandin:
Sorry for not following up on this pull request. I would of course be happy if standardized-audio-context proofs to be useful for Tone.js. Let me know if there is anything I need to do to make it work.

tambien:
@chrisguttandin thanks! i will let you know how it goes

Issue #405, opened 2018-11-19
Is Tone.Sampler unavailable in Tone.Offline?


yos1up:
I want to record a melody using **Tone.Offline**, and then play it with **Tone.Player**.

Using **Tone.Synth** in **Tone.Offline** works fine ( `play3()` below), but using **Tone.Sampler** in **Tone.Offline** does not work ( `play4()` below, no sound heard). How do I use **Tone.Sampler** in **Tone.Offline**?

```js
<!DOCTYPE html>
<html>
<head>
    <title></title>
</head>
<body onload="init()">
<button onclick="play3();">Play3</button>
<button onclick="play4();">Play4</button>
</body>
<script type="text/javascript" src="https://tonejs.github.io/build/Tone.min.js"></script>
<script type="text/javascript">
var player;
function init(){
    player = new Tone.Player().toMaster();
}
function play3(){ // Offline-recording then playing [synth]. Works fine.
    Tone.Offline(function(Transport){
        var synth = new Tone.Synth();
        synth.toMaster();
        synth.triggerAttackRelease('A4', '4n');      
    }, 4).then(function(buffer){
        player.buffer = buffer;
        player.start();
    });
}
function play4(){ // Offline-recording then playing [sampler]. Does not work.
    Tone.Offline(function(Transport){
        console.log(Tone.Master); // -> OfflineContext
        var offlineMaster = Tone.Master;
        var offlineSampler = new Tone.Sampler({
            "C2" : "C2.wav",
            "E2" : "E2.wav",
            "Ab2" : "Ab2.wav",
            "C3" : "C3.wav",
            "E3" : "E3.wav",
            "Ab3" : "Ab3.wav",
            "C4" : "C4.wav",
            "E4" : "E4.wav",
            "Ab4" : "Ab4.wav",
            "C5" : "C5.wav",
            "E5" : "E5.wav",
            "Ab5" : "Ab5.wav",
            "C6" : "C6.wav",
        }, {
            'release': 1,
            'onload': function(){
                console.log(Tone.Master); // -> Context not offline
                // offlineSampler.toMaster(); // fails (nothing shown, nothing heard)

                offlineSampler.connect(offlineMaster); // succeeds

                console.log('connected to Master!'); // shown

                offlineSampler.triggerAttackRelease('A4', '4n'); // still fails (nothing shown, nothing heard)

                console.log('triggered!'); // not shown
            },
            'baseUrl': "http://localhost:8686/"
        });
[...]

Issue #368, opened 2018-07-10
Suggestion: Tone.Offline might benefit from ability to control when the rendering process starts


gdenisov:
When using Sampler and Players in the Tone.Offline callback I am running into an issue that by the time Sampler or Player nodes have been loaded rendering is already done. It might be useful for offline callback to look something like this:

```
Tone.Offline((Transport, startRender) => {
  new Sampler({
    "C3" : "path/to/C3.mp3",
    "D#3" : "path/to/Dsharp3.mp3",
    "F#3" : "path/to/Fsharp3.mp3",
    "A3" : "path/to/A3.mp3",
   }, () => {
    // ... schedule events
    startRender()
  })
})
```

Of course we could preload buffers outside of Tone.Offline call and then pass them in to the sampler in the callback, but this becomes increasingly more difficult when the node creation is abstracted in another external function

tambien:
Thanks for the suggestion, i like your proposed solution. 

There is currently an undocumented way to do this, but i'm not totally satisfied with the API yet. Currently you can return a promise within Tone.Offline, and the context will wait until that promise is resolved before starting to render.

rodw:
@tambien can you give an example of what you mean by "return a promise within Tone.Offline, and the context will wait until that promise is resolved before starting to render"?

rodw:
The attached HTML document (added with a `.txt` extension because GitHub won't let me upload it otherwise) demonstrates the problem, or at least the way I'm trying to invoke this.

When opened, the HTML document has four links:

- Basic Synth (online) - plays some notes using Tone.Synth directly
- Sampler (online) - plays some notes using Tone.Sampler directly
- Basic Synth (offline) - plays some notes using Tone.Synth via Tone.Offline (then plays the generated buffer)
- Sampler (offine) - plays some notes using Tone.Sampler via Tone.Offline (then plays the generated buffer), or at least tries to
 
The first three links seem to work fine and can be repeated an arbitrary number of times.

The fourth (offline sampler) doesn't work as-is and I'm not able to get it to work in any variation I've tried.  In the current setup I'm getting "Error connecting to node: [object GainNode]" when I try to call `sampler.triggerAttackRelease`.

That error (and some other debugging output) can be seen in the console log when that link is hit. 

@tambien Am I doing this incorrectly? Should I open a new issue with this example? 

[offline-sampler.html.txt](https://github.com/Tonejs/Tone.js/files/2506792/offline-sampler.html.txt)

[...]

PR #441, opened 2019-02-06
Should OfflineContext be available for the entire duration of an async callback?

salomvary:
The problem is, the second test case I added ("does not restore Tone.context during asynchronous calls") is failing.

The other problem is, it's not easy to fix, the easy attempt will break the third test case I added ("restores Tone.context immediately").

Use case: `fetch` media files for offline rendering and use `Tone.context.decodeAudioData` to decode them.

I understand there are various workarounds (fetch *before* the `Tone.Offline` call, pass down a reference to the `OfflineContext` instance, etc.) but my first attempt wrongly assuming `Tone.context` is an `OfflineContext` even in async calls ended up in cryptic exceptions (esp. in Firefox) and quite a bit of headscratching.

Possible solutions I see:
- Breaking API change to not modify `Tone.context` at all and expose it as a callback argument
- A warning logged when attempting what I did in "does not restore Tone.context during asynchronous calls" test case
- A warning in the documentation and suggested workaround

@tambien What do you think?

### DIFF
--- a/test/core/Offline.js
+++ b/test/core/Offline.js
@@ -1,5 +1,7 @@
 import Test from "helper/Test";
 import Offline from "Tone/core/Offline";
+import Context from "Tone/core/Context";
+import OfflineContext from "Tone/core/OfflineContext";
 import Transport from "Tone/core/Transport";
 import Oscillator from "Tone/source/Oscillator";
 import Tone from "Tone/core/Tone";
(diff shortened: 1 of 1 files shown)


tambien:
yeah this is a tricky one. i've been thinking about this for a while since it's also related to #368 and #405. I think the thing that needs to happen is that you should be able to create a Tone object which is bound to the context the entire time, instead of temporarily switching the global context. 

So i would imagine an API being something like `const OfflineTone = Tone.fromContext(new OfflineAudioContext(params))` which would return a Tone object with the same interface, but the context is bound. This would give a lot more flexibility with how and when it would render. 

I've been laying the ground work for this for a little while, but it would be a bit of a refactor to get that in there.

salomvary:
> you should be able to create a Tone object which is bound to the context the entire time

👍 

Anyway, closing this PR as it is just a byproduct of me saying "I can quickly fix this!" and then "Oh wait...".  It was still useful to illustrate an issue...

salomvary:
Btw this also means it is currently not possible to use `Context` and `OfflineContext` the same time.

PR #457, opened 2019-03-11
Fixed 'notes' parameter in Sampler docs.

vasiliismirnov:
Fixed 'notes' parameter in Sampler methods to point that several notes can be played.


tambien:
@vasiliismirnov thank you!

Issue #451, opened 2019-03-03
Add super-simple HTML5 example?


RobJacobson:
I'm a complete newbie to Tone.js.  I'm having some basic problems with following the tutorials on the website.  E.g., the examples on the website all require external scripts, and I'm not sure what they do.  I'm also finding it difficult to import Tone.js into a modern "app.js" file under ES6+.  (I've read the docs but they don't answer these questions.)

Would it be possible to create a "stupid simple example" in pure HTML5/ES6+, using the CDN libaries?  It wouldn't need to be more complicated than the following:

1.  Show a button in HTML that says "click me," or the equivalent.

2.  Have a separate "app.js" file that responds to the click and plays a single note using Tone.js (using external linking to the CDN libraries).

Thank you.

isaacpearl:
Just created a pull request to add this!

RobJacobson:
Thank you!  Exactly what I needed!

PR #465, opened 2019-03-20
Added simple HTML5 example to examples folder

isaacpearl:
Created basic HTML5 example as per https://github.com/Tonejs/Tone.js/issues/451. This is my first time opening a pull request, so please let me know if I need to do things differently and/or rerequest! 

I assumed that I shouldn't use the tonejs-ui.js script for the sake of simplicity, but if there's some reformatting that should be done, I'm happy to do it.

:)


tambien:
Thank you! you've got two extra .swp files in there. Could you please remove those, and also put the script code inside the html files in a <script> tag. I think that'll also help simplify things by not referencing an external file.

isaacpearl:
@tambien how does this look?

tambien:
@isaacpearl thanks!

PR #469, opened 2019-03-26
fixes #461

### LINKED ISSUE #461: Disconnect from PanVol Not Working
Unable to disconnect from PanVol

Tone.js v13.8.1

example:
https://jsfiddle.net/timboie/bhemrv2d

tboie:
file change history:

https://github.com/Tonejs/Tone.js/commit/eb8bdc0f3741e155e683504f1eece5d2e90e37b2#diff-54c4c358f8061be4571a2e1a0ac2deefL361

### DIFF
--- a/Tone/core/Tone.js
+++ b/Tone/core/Tone.js
@@ -365,7 +365,7 @@ Tone.connect = function(srcNode, dstNode, outputNumber, inputNumber){
 Tone.disconnect = function(srcNode, dstNode, outputNumber, inputNumber){
 	if (dstNode){
 		//resolve the input of the dstNode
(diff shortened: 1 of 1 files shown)


tboie:
This fix ok?  Does it just need to be re-checked?

tambien:
it did not pass the tests which makes me think that it might not be the solution and the type of failure makes me think that `while` loops infinitely.  i have not had a chance to investigate further.

tboie:
Ok , looking into this

tboie:
The catch block is causing the timeout when a disconnect error is thrown.  Moving it outside the while block.

tboie:
Closing. different pull request has been submitted

Issue #471, opened 2019-04-04
Panner object parameter not supported


tboie:
Object parameter not supported.  This functionality is consistent throughout the library.

 * Tone.js v13.8.5 * 

Demo:
https://jsfiddle.net/timboie/mu28xgeq/

tambien:
Thank you! closed by #472

PR #472, opened 2019-04-04
#471

tboie:
Panner construction with options object


tambien:
this is great! thank you

PR #474, opened 2019-04-11
README for examples directory

colinmorris:
It took me a fairly long time to figure out what was going on with the web components used in the examples. I tried grepping for the component names like `tone-example`, `tone-fft`, but that just told me they were defined in some blob of minified js at `examples/js/tonejs-ui.js`. Hopefully this clarifies for future readers.


tambien:
thanks!!

PR #476, opened 2019-04-11
Correct description for Tone.FFT

colinmorris:
Previous class description looks like it was mistakenly copy-pasted from `Tone.Waveform`.


tambien:
thank you!

PR #477, opened 2019-04-12
Improved explanation of size in FFT/Analyser docs

### LINKED ISSUE #475: Documented size limits for FFT/Analyser are incorrect
[Tone.FFT](https://tonejs.github.io/docs/r13/FFT) and [Tone.Analyser](https://tonejs.github.io/docs/r13/Analyser) say that size should be "a power of two in the range 32 to 32768". The actual allowable range is 16-16384. If you try to set it to 32768, you'll get an error like: `Uncaught DOMException: Failed to set the 'fftSize' property on 'AnalyserNode': The FFT size provided (65536) is outside the range [32, 32768].`

32-32768 is the allowable size range for [AnalyserNode](https://webaudio.github.io/web-audio-api/#analysernode), but the size value that `Tone.Analyser` passes to its `AnalyserNode` is double the input size provided.

A couple options:
1. Stop doubling input size when setting fftSize
2. Change docs to reflect actual limits

Presumably number 2 would be preferable since it's a non-breaking change?

Could also consider renaming the parameter from `size` to something like `bins`. (I think this would make the semantics more clear, and make it more apparent why it's doubled to get fftSize).

colinmorris:
Fixes #475

### DIFF
--- a/Tone/component/Analyser.js
+++ b/Tone/component/Analyser.js
@@ -9,7 +9,7 @@ import "../shim/AnalyserNode";
  *  @extends {Tone.AudioNode}
  *  @param {String=} type The return type of the analysis, either "fft", or "waveform".
  *  @param {Number=} size The size of the FFT. Value must be a power of
- *                       two in the range 32 to 32768.
+ *                       two in the range 16 to 16384.
  */
 Tone.Analyser = function(){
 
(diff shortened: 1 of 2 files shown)


tambien:
@colinmorris thank you!

Issue #473, opened 2019-04-10
issue on offline export sequence/part


zfan40:
I noticed there have been couple of issues on offline export, and I apologize if mine is a duplicate.
the demo you gave works pretty well, well I run into a problem using Tone.part and Tone.Sequence in Offline context
 In my demo below, the offline context seems abandon all the notes after 2 second.
https://jsfiddle.net/gztLy5cn/2/
I did find a workaround though by triggering notes one by one without Part/Sequence.

tboie:
try removing part.loop = false in the Offline func for now

tboie:
https://github.com/Tonejs/Tone.js/blob/dev/Tone/event/Part.js#L360

Look like testLoopBoundries is called when loop is set for all conditions:

https://github.com/Tonejs/Tone.js/blob/dev/Tone/event/Part.js#L418

PR #478, opened 2019-04-12
#473

tboie:
Fixes Part loop toggle


tambien:
@tboie thank you!

fixes #473

PR #481, opened 2019-04-21
Disconnect from nested input nodes

tboie:
#461


tboie:
Just realized a potential problem if a nested input is an array.  Will change soon

PR #488, opened 2019-05-13
Change const to var for ES5 compatibility

### LINKED ISSUE #485: The entry file contains keyword `const`, which is not allowed in es5
I found the entry file `build/Tone.js` contains the reserved keyword `const` which is not allowed in es5.
image

stephlow:
Simple change, fixes #485.


tambien:
thanks!

PR #503, opened 2019-06-20
[wip] convert Filter to .ts

garrensmith:
This is an initial start on converting the filter.js to typescript. Its still a huge WIP but I thought I would start a draft PR so that no one else starts working on the filter.

### DIFF
--- a/Tone/component/filter/Filter.ts
+++ b/Tone/component/filter/Filter.ts
@@ -1,0 +1,200 @@
+import {ToneAudioNode, ToneAudioNodeOptions, InputNode, OutputNode} from '../../core/context/ToneAudioNode';
+import { optionsFromArguments } from "Tone/core/util/Defaults";
+import { Signal } from 'Tone/signal/Signal';
+import { readOnly, writable } from 'Tone/core/util/Interface';
+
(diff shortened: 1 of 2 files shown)


tambien:
this is a great start!

Tone/component/filter/Filter.ts
```diff
@@ -0,0 +1,76 @@
...
+
+	name = "Filter";
+	input = InputNode;
+
+	constructor(type?: FilterOptions);
+	constructor(options?: Partial<FilterOptions>);
+	constructor() {
+		super(optionsFromArguments(Filter.getDefaults(), arguments, ["frequency", "type", "rolloff"]));
+
+		this.createInsOuts(1,1);
```
tambien: i think no need for this function. check out [Analyser.ts](https://github.com/Tonejs/Tone.js/blob/typescript/Tone/component/analysis/Analyser.ts) which has a similar thin wrapping of a native node. it just declares the filter as the input and output node.

garrensmith:
@tambien awesome. Thanks for the feedback. I've converted all of it along with the test. Its definitely not perfect yet. I'm a little stuck on the types, I think my knowledge of typescript breaksdown here https://github.com/Tonejs/Tone.js/pull/503/files#diff-6aed955877c105e29b449324d18737ebR45

How do we handle that as a `filtertype` and a number?

Issue #494, opened 2019-05-27
Part plays twice, when Part.loop = 1


meszaros-lajos-gyorgy:
Hi!

_I made another jsfiddle to illustrate the issue: https://jsfiddle.net/lmeszaros/yL1zx78u/39/_

I have `Part.loop` set to `1` and `Transport.loop` set to `false`, but Part will loop as if Part.loop was set to 2.

jaredellison:
Hi! Just discovered this library and started looking at a few open issues to get familiar.	

This loop logic appears to be implemented in [`Tone.Event.prototype._rescheduleEvents`](https://github.com/jaredellison/Tone.js/blob/305e716cde80faaf8600e5e500b94f551daa1f7d/Tone/event/Event.js#L139)  and the question seems to be over what happens when `loop` is 0, 1, or greater than 1.

I'm with you @meszaros-lajos-gyorgy, it seems like the sequence should play once when `loop === 1` and twice when `loop === 2`.  

Currently, if `loop ` is false or 0, single events are scheduled and if `loop`  is 1 or greater, repeating events are scheduled (even if it's one repeat).

I drafted a pull request (#508) where `Tone.Event.prototype._rescheduleEvents` handles 0 and 1 loops in the same way - the sequence plays only once and there is no change to the functionality when `loop` is true, false or greater than 1.

This prompts the questions: 

- Should `loop === 0` actually mean that no events are scheduled? 
- Would it be helpful to add a [test](https://github.com/Tonejs/Tone.js/blob/305e716cde80faaf8600e5e500b94f551daa1f7d/test/event/Event.js#L293) for this behavior around 0 or 1 loops?

meszaros-lajos-gyorgy:
While we are at checking on the loop property, I would have a suggestion:

I don't know the history of the `loop` property for the `Part` and `Transport` classes, but from a developer's perspective I think it would be clearer, if it would only accept integers. Does `loop === false` mean `loop === 0`? Does `loop === true` mean `loop === Number.MAX_SAFE_INTEGER`?

Or if the boolean values are more reasonable for the property, then the quantified value should be set via a different property, for example `loopFor`.

tambien:
@jaredellison thanks for the PR! it'd be great to include a test like the one you linked to validate your PR. 

[...]

PR #508, opened 2019-07-05
Refactor _rescheduleEvents to prevent repeating events when loop = 1

jaredellison:
Refactoring looping logic in reference to #494


tambien:
thanks!

tambien:
fixes #494

Issue #513, opened 2019-07-18
Signal doesn't work with PitchShifter pitch


Foaly:
I think I discovered a bug in Tone.js. I was trying to get some help on the forum, but I didn't get a response, so I will try again here.

I want to change the pitch of the Pitchshifter over time. Unfortunatelly I am unable to acchive this. It seems to me as if the pitch member of the PitchShifter is not an AudioParam and can thus not be used with Signal. To me this seems like a bug. But I'm not quite sure what to do about it. Is there another way to interpolate the pitch smoothly? Thank you so much in advance!

Here is my code:
```javascript
  var shifter = new Tone.PitchShift(0).toMaster();

  var pitchShiftSignal = new Tone.Signal(0, Tone.Type.Interval);
  pitchShiftSignal.connect(shifter.pitch);

  var player = new Tone.Player("./sounds/cello/cello-a2.wav").connect(shifter);
  player.loop = true;

  $("#playButton").click(function() {
    player.start();

    pitchShiftSignal.linearRampTo(2, 2);
  });
```
I'm getting this error:
```
Error: error connecting to node: 0
TypeError: Argument 1 is not valid for any of the 2-argument overloads of AudioNode.connect.
```

tambien:
This is not necessarily a bug, but related to how the PitchShifter is implemented. The pitch attribute is a number and not an AudioParam so it can't be scheduled and connected to in the same way as an AudioParam. I described more fully the reason for this limitation [here](https://github.com/Tonejs/Tone.js/issues/509#issuecomment-509236749).

There is currently no easy way around that other than to incrementally update the `pitch` attribute on a setInterval. 

It seems possible to reimplement Tone.Scale with a Tone.Add and Tone.Multiply instead of a WaveShaper which could propagate all the way up to the PitchShifter enabling it to have a signal-rate pitch attribute.

Foaly:
Thank you for the fast answer and explanation!
My javascript is a bit rusty so maybe I am missing something but looking at `Scale.js` I can only see Add and Multiply being used. So maybe it is possible to change `pitch` into an AudioParam?

tambien:
you're right! it's currently implemented with a Multiply and Add, i guess the part that would need to be updated it making max/min signal-rate instead of just numbers.

Foaly:
Ah thats fantastic news!
Do you think it's worthwhile to reopen this ticket?
[...]

PR #514, opened 2019-07-20
Refactored Scale.js to work at audio rate.

Foaly:
Hi there!
I reworked `Scale.js` to work at audio rate using only Signal objects. I was sticking to the tests while refactoring, which are all passing now. I hope there were no side effects I missed.
I don't have a lot of experience developing JavaScript libraries and this is my first time playing with web audio, so I can't really judge, if this change is a performance improvement or how it affects other object depending on `Scale`.
Any feedback is very welcome :)

### DIFF
--- a/Tone/signal/Scale.js
+++ b/Tone/signal/Scale.js
@@ -1,4 +1,5 @@
 import Tone from "../core/Tone";
+import "../signal/Subtract";
 import "../signal/Add";
 import "../signal/Multiply";
 import "../signal/Signal";
@@ -10,8 +11,8 @@ import "../signal/Signal";
  *
  *  @constructor
  *  @extends {Tone.SignalBase}
- *  @param {number} [outputMin=0] The output value when the input is 0. 
- *  @param {number} [outputMax=1]	The output value when the input is 1. 
+ *  @param {number} [outputMin=0] The output value when the input is 0.
+ *  @param {number} [outputMax=1] The output value when the input is 1.
  *  @example
  * var scale = new Tone.Scale(50, 100);
  * var signal = new Tone.Signal(0.5).connect(scale);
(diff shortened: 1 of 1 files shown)


Foaly:
This was meant to fix #513. While all tests are passing now and there are no more console errors, I still can't modulate the `PitchShift` `pitch` property. Is there anything else down the line (in `LFO` or `PitchShift`) that I have to change to make this change affect the `pitch` property? I couldn't see anything obvious.

Here is the code I used for testing:
```javascript
  shifter = new Tone.PitchShift(0).toMaster();

  pitchShiftSignal = new Tone.Signal(0, Tone.Type.Interval);
  pitchShiftSignal.connect(shifter.pitch);

  player = new Tone.Player("sound.wav").connect(shifter);
  player.loop = true;

  $("#playButton").click(function() {
    player.start();

    pitchShiftSignal.linearRampTo(2, 2);
  });
```

tambien:
nicely done! but i'm really sorry, i may have led you astray. Scale min/max being signal-rate is one of the reasons that PitchShift.pitch isn't a Signal parameter, but [looking at the code](https://github.com/Tonejs/Tone.js/blob/dev/Tone/effect/PitchShift.js#L161) again now, there's a few other factors as well. The logic of setting the lfo min/max would also need to be there in order to make pitch a signal-rate property. All of that logic would also need to be implemented as signal-rate parameters. 

Again, this logic might also be best implemented in an AudioWorklet...

Foaly:
Yes... I realized that pretty quickly after implementing this and taking another look at `PitchShift` with my gained understanding on the WebAudio API. From looking at it seems that the algorithm is not really set up to be implemented with just Signals (like the lfo min/max switching). 
But I wasn't really happy with the modulation effect the PitchShift introduced as a side effect at higher pitch values anyway, so I am thinking of maybe implementing a Frequency Shifter using a SSB modulator.

Anyways if this PR is still useful to you and does not impact performance, feel free to merge.

tambien:
Going to close this for now. Thanks for the PR! sorry again for leading you down the wrong route. Don't think it's worth the performance hit of the added components without any use for it currently. will reopen if it turns out to be helpful.

PR #517, opened 2019-07-28
Fixing released-sustained notes on Tone.Sampler

lidiovargas:
When playing in sampled piano controlled by sustain (CC64), when the pedal is down (CC64=127), and you play the same note twice (or more), the original code removes only the first `BufferSource` in `this._activeSources` array. Is expected that all the BufferSources of the list for the relased midi-note need to be stopped, and not only the first one. 

The original `Tone.Sampler.prototype.triggerRelease`:
```
var source = this._activeSources[midi].shift();
time = this.toSeconds(time);
source.stop(time);
```

The proposed changes:
```
time = this.toSeconds(time);
while (this._activeSources[midi].length){
	var source = this._activeSources[midi].shift();
	source.stop(time);
}
```
Of course, depends in how your instrument object will implement the Tone.Sampler class. But I think the responsability is of the Tone.Sampler class instead of the instrument implementation, in this case. 

This behavior was tested on https://pianoplayer.app demo.

NOTES: your library is awesome, men. Thank you for that.


lidiovargas:
i will change the code to pass the travis tests, sorry by that.

tambien:
@lidiovargas looks like it may have been just a simple linting error.

lidiovargas:
@tambien it's a missing semicolon. But after passing this lint error, I changed the `forEach`  to `while` loop, to pass the  `test/instrument/Sampler.js:it("matches a file...")`. The tests are passing, if you want to appreciate it now.

PR #521, opened 2019-08-01
FrequencyShifter

Foaly:
Hi there!
Like I mentioned in #514 for a project I am working on I was in need of a frequency shifter that is able to change the pitch at audiorate. Since it seemed difficult to change the implementation of the existing `PitchShift`, I went ahead and implemented a new frequency shifter. It implements a single side band modulator with phase canceling using a Hilbert Transform to isolate the single side band. The Hilbert Transform is implemented very efficiently using two allpass filterbanks. I wrote a seperate component `PhaseShiftAllpass` with tests for this. You can read an explanation of how the SSB modulator algorithm works in this nice write up: https://larzeitlin.github.io/RMFS/
Since `FrequencyShifter` is a not a pitch shifter, but a frequency shifter all frequencies of the input signal are getting shifted by the same amount, destroying harmonic overtone series and changing the timbre, resulting in the classic ring modulator sound.
Anyway I think this would be a nice addition to this library :)


tambien:
this is awesome! thank you! i will take a listen as soon as i've got a chance

Foaly:
Thanks :) Cool! 
Yeah there is some aliasing towards the higher end, especially if the input contains a lot of high frequencies. In order to get rid of it the input would need to be upsampled by a factor of 2 and then low pass filtered before mixing and downsampled at the output. I'm not really sure how to do up- and downsampling in Tone.js or WebAudio. Maybe you know?

tambien:
@Foaly AFAIK it's not possible to upsample and then downsample in a stream in Web Audio. The only way i've seen is using the OfflineAudioContext, but that's not gonna help in this situation.

Foaly:
Hmm thats unfortunate. I guess then there is not much we can do about the aliasing for now... I will add a note explaining it.
I looked into the failing test on safari, but didn't really understand why it happens. It looks like `Tone.context` is null in the Offline callback. Do you have an idea what could cause this? Is it maybe a click-something-to-activate-the-audio-context thing?

tambien:
looks like you’re using the IIR filter which safari does not yet support. is it possible to do the filtering step with the biquad filter?

Foaly:
For some reason the webaudio biquads don't support setting the filter coefficients directly. This is needed for both the allpass filters and the delay, so unfortunately I think using a shim is the only option. Is this something you could help me with? I have no idea how they work.

tambien:
I'll merge this PR into the typescript branch instead of 'dev'. i'm planning on adding shims from [standardized-audio-context](https://github.com/chrisguttandin/standardized-audio-context) which should cover this situation, but i won't be including that in the 'dev' branch for now.

Foaly:
Sure, if that fits the development better feel free to merge it to the typescript branch. Whats the plan/status on it?
I will add some more documentation on the aliasing when I find a chance in the next couple of days. Other than that its good to go from my side :)

Foaly:
Cool thanks :)

PR #522, opened 2019-08-03
converting Add to typescript

this-fifo:
Hey @tambien , looks like Add is one of the first files I'm gonna need to touch in order to get up all the way to converting MonoSynth so I decided to start small

I tried basically following what you did on Subtract.ts, looks like it's currently failing for 2 tests when handling I/O and when adding two signals but I am having trouble getting these to work, let me know if you can help a bit :)

### DIFF
--- a/Tone/signal/Add.test.ts
+++ b/Tone/signal/Add.test.ts
@@ -1,0 +1,51 @@
+import { BasicTests } from "test/helper/Basic";
+import { connectFrom, connectTo } from "test/helper/Connect";
+import { ConstantOutput } from "test/helper/ConstantOutput";
+import { Add } from "./Add";
+import { Signal } from "./Signal";
+
+describe("Add", () => {
+
+	BasicTests(Add);
+
+	context("Addition", () => {
+
+		it("handles input and output connections", () => {
+			const add = new Add();
+			connectFrom().connect(add);
+			connectFrom().connect(add, 0);
+			connectFrom().connect(add, 1);
+			add.connect(connectTo());
+			add.dispose();
+		});
+
+		it("correctly sums a signal and a number", () => {
+			return ConstantOutput(() => {
+				const signal = new Signal(0);
+				const adder = new Add(3);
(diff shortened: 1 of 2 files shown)


tambien:
thank you! this was really helpful! 

A reason the tests weren't passing, is because i'm moving away from `input` as an array and moving in favor of just having multiple named inputs. so i added another input named `addend` which turns out to be the technical name of the thing which is added, and a second signal can be connected directly to that. 

So for example

OLD:
```
const sigA = new Signal(1);
const sigB = new Signal(4);
const adder = new Add();
sigA.connect(adder, 0, 0);
sigB.connect(adder, 0, 1);
```

NEW
```
const sigA = new Signal(1);
const sigB = new Signal(4);
const adder = new Add();
sigA.connect(adder);
sigB.connect(adder.addend);
```

I think it's more clear and readable. and simplifies connect/disconnect logic greatly

this-fifo:
Ha! That makes a lot more sense now, I was trying to understand why Subtract had the subthraend param which wasn’t there before and this explains it well.

Thanks for taking a look into that, I’m gonna continue up the chain now, hopefully the more I do this the better the momentum goes

On Aug 4, 2019, at 10:22 AM, Yotam Mann <notifications@github.com<mailto:notifications@github.com>> wrote:


thank you! this was really helpful!

A reason the tests weren't passing, is because i'm moving away from input as an array and moving in favor of just having multiple named inputs. so i added another input named addend which turns out to be the technical name of the thing which is added, and a second signal can be connected directly to that.

So for example

OLD:

const sigA = new Signal(1);
const sigB = new Signal(4);
const adder = new Add();
sigA.connect(adder, 0, 0);
sigB.connect(adder, 0, 1);


NEW

const sigA = new Signal(1);
const sigB = new Signal(4);
const adder = new Add();
sigA.connect(adder);
sigB.connect(adder.addend);


I think it's more clear and readable. and simplifies connect/disconnect logic greatly

PR #523, opened 2019-08-04
Typescript

this-fifo:
@tambien this time I'm converting Scale to typescript, for some reason I can't seem to be getting scaling from inputs to work, might be related to how I'm connecting the signals or how I'm setting the value, gonna need some help here again

### DIFF
--- a/Tone/signal/Scale.ts
+++ b/Tone/signal/Scale.ts
@@ -5,8 +5,7 @@ import { Multiply } from "./Multiply";
 import { Signal } from "./Signal";
 import { SignalOperator } from "./SignalOperator";
 
-export interface ScaleOptions<Type> extends ToneAudioNodeOptions {
-	value: Type;
+export interface ScaleOptions extends ToneAudioNodeOptions {
 	outputMin: number;
 	outputMax: number;
 }
@@ -25,7 +24,7 @@ export interface ScaleOptions<Type> extends ToneAudioNodeOptions {
  * var signal = new Signal(0.5).connect(scale);
  * //the output of scale equals 75
  */
-export class Scale extends SignalOperator<ScaleOptions<number>> {
+export class Scale extends SignalOperator<ScaleOptions> {
 
 	readonly name: string = "Scale";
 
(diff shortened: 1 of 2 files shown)


Tone/signal/Scale.ts
```diff
@@ -0,0 +1,104 @@
...
+	// tslint:disable-next-line: unified-signatures
+	constructor(value?: number);
+	constructor() {
+		super(Object.assign(optionsFromArguments(Scale.getDefaults(), arguments, ["value"])));
+
+		const options = optionsFromArguments(Scale.getDefaults(), arguments, ["outputMin", "outputMax"]);
+		this._outputMin = options.outputMin;
+		this._outputMax = options.outputMax;
+
+		this.connect(this.output);
```
tambien: i think there's an issue with this line. it should probably be `this.input.connect(this.output)`

Tone/signal/Scale.ts
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	static getDefaults(): ScaleOptions<number> {
+		return Object.assign(Signal.getDefaults(), {
+			outputMax: 1,
+			outputMin: 0,
+		});
+	}
+
+	get getMin(): number {
```
tambien: there seems to also be a typo here. should be `get min()` and `set min(min)` instead of `get getMin()` and `set setMin()`

Tone/signal/Scale.ts
this-fifo: I tried that, still getting the tests to fail but the error seems more meaningful now, I'm getting
```AssertionError: expected -9 to be close to 8 +/- 0.01 ```
```AssertionError: expected 10.5 to be close to 15 +/- 0.01 ```

this-fifo: found the issue, I was using the incorrect property inside _setRange

tambien:
thanks for the PR! i updated some of the naming and docs slightly. looks like you forgot to port over the docs for get/set min/max, so i added those. also renamed outputMin/Max to min/max since it's what those attributes are called on the object and it seems clear if they are totally consistent

this-fifo:
awesome, thanks for reviewing and helping!

I am starting to understand more your approach with this refactoring, I'm gonna try being more consistent in future PRs and keep an eye for those things as well

tambien:
really appreciate the help! so thanks again!

PR #543, opened 2019-09-16
convert Pow to typescript

this-fifo:
Hey @tambien , I'm running into issues getting the set new exponent test to work, let me know if you have any ideas. After I get Pow to work I plan on looking into converting ScaledEnvelope which requires Pow


this-fifo:
I noticed the problem with set was that it was using ```value``` instead of ```exponent``` but now looking at the other files I think I should refactor that to use ```Param``` instead

tambien:
thanks for your work on this. I would avoid introducing `exponent` as a member and keep it to `value`. I think in this case you want to extend SignalOperator (formerly called SignalBase) instead of Signal. Since most of the properties i don't think are necessary/possible in terms of scheduling the exponent value. which can't be scheduled like a Signal or Param because it is implemented using a waveshaper function.

this-fifo:
I was trying to follow the approach on ```Multiply```, it does look like I can't use signal or param due to the waveshaper function, but I got confused trying to use SignalOperator because ```ToneAudioNodeOptions``` doesn't have a ```value``` so I always get ```NaN``` on my tests when I try using that

tambien:
I think a better example to use than Multiply is AudioToGain or GainToAudio.ts. Those both extend the SignalOperator class which is meant as a basic signal processor. Multiply, Add, Subtract all include schedulable values which this does not. 

I think i see why you're getting this NaN as your value in the tests. I'll add some annotation to hopefully make it a little more clear.

Tone/signal/Pow.ts
```diff
@@ -0,0 +1,85 @@
...
+ *         must be AudioRange.
+ *
+ *  @constructor
+ *  @param exp The exponent to apply to the incoming signal, must be at least 2. 
+ *  @example
+ * var pow = new Pow(2);
+ * var sig = new Signal(0.5).connect(pow);
+ * //output of pow is 0.25. 
+ */
+export class Pow extends Signal<number> {
```
tambien: This Should extend [SignalOperator](https://github.com/Tonejs/Tone.js/blob/typescript/Tone/signal/SignalOperator.ts)

and pass in PowOptions as the generic argument to SignalOperator. This will let the get/set functions know what the defaults interface looks like

Tone/signal/Pow.ts
```diff
@@ -0,0 +1,85 @@
...
+	
+	readonly name: string = "Pow";
+
+	/**
+	 * Indicates if the value should be overridden on connection
+	 */
+	readonly override = false;
```
tambien: This probably isn't necessary. Override is a flag which indicates if a value should be overwritten when it is connected to, but since it's not a schedulable value like Signal or Multiply, there is nothing to override. the signal is just processed as it passes through the unit.

Tone/signal/Pow.ts
```diff
@@ -0,0 +1,85 @@
+import { Signal, SignalOptions } from "./Signal";
+import { WaveShaper, WaveShaperMappingFn } from "./WaveShaper";
+import { optionsFromArguments } from "Tone";
+
```
tambien: Add an interface which extends ToneAudioNodeOptions called PowOptions which might look something like this: 

```typescript
interface PowOptions extends ToneAudioNodeOptions {
  value: number;
}
```

Tone/signal/Pow.ts
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	 * @param value Constant exponent value to use
+	 */
+	constructor(value?: number);
+	constructor(options?: Partial<SignalOptions<number>>);
+	constructor() {
+		super(Object.assign(optionsFromArguments(Pow.getDefaults(), arguments, ["value"])));
+
```
tambien: This part is kind of weird, but you'll actually call a nearly identical line twice here and it'll look something like this: 

```typescript
super(optionsFromArguments(Pow.getDefaults(), arguments, ["value"]));
const options = optionsFromArguments(Pow.getDefaults(), arguments, ["value"]);
```

That second `options` can be used to get the passed in value.

Tone/signal/Pow.ts
```diff
@@ -0,0 +1,85 @@
...
+		this.exponent = this.value;
```
tambien: Here you can get the passed in value by using something like

```typescript
this.value = options.value
```

Also change the variable from `exponent` to `value` to be more consistent with the previous implementation. 

This will fix your NaN issue since it'll be setting this.value to the default options.value or the passed in value.

Tone/signal/Pow.ts
```diff
@@ -0,0 +1,85 @@
...
+	 * the function which maps the waveshaper
+	 * @param exp exponent value
+	 */
+	private _expFunc(exp: number): WaveShaperMappingFn {
+		return (val: number) => {
+			return Math.pow(Math.abs(val), exp);
+		};
+	}
+
+	get exponent(): number {
```
tambien: This should read be get/set `value` instead of `exponent`

Tone/signal/Pow.ts
```diff
@@ -0,0 +1,85 @@
...
+	 * Indicates if the value should be overridden on connection
+	 */
+	readonly override = false;
+	
+	/**
+	 * The exponent
+	 */
+	private _exponent!: number;
+	
+	private _exponentScaler = new WaveShaper({
```
tambien: As a small optimization you can move this class constructor into the `constructor()` function which will save setting the exponent function twice, once when it's created and another time when the exponent is set.

Tone/signal/Pow.ts
this-fifo: This makes a lot more sense now, I was going towards that route, I basically forgot I could just extend ToneAudioNodeOptions to include the value for that class!

Tone/signal/Pow.ts
this-fifo: Thanks for the explanation, I did not realize that at first but it does make sense looking at it again

Tone/signal/Pow.ts
this-fifo: I noticed that on other classes, I thought it was weird too and then I realized it was the only possibly way since the super call is required to be the first statement

tambien:
thanks for the PR!

PR #548, opened 2019-09-29
catch errors thrown by playbackRate AudioParam in Safari

chrisguttandin:
This pull request is meant to fix the problem that the playbackRate AudioParam doesn't support connecting something to it and standardized-audio-context throws an error when connecting something to it.

### DIFF
--- a/Tone/core/context/Param.ts
+++ b/Tone/core/context/Param.ts
@@ -115,7 +115,13 @@ export class Param<Type extends Unit = number>
 		this.input = this.context.createGain();
 		// initialize
 		this._param = options.param;
-		this.input.connect(this._param);
+		try {
+			this.input.connect(this._param);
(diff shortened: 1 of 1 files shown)


tambien:
thanks @chrisguttandin, I ended up going a slightly different direction on this which is in cases when the underlying parameter is swappable and therefore necessitates a gain node be connected to it, then i [mark it as swappable](https://github.com/Tonejs/Tone.js/blob/typescript/Tone/core/context/Param.ts#L111-L119) in the constructor. Simplifies the code and produces fewer extra gain nodes since only parameters which need to be replaced at some point will construct that GainNode

Going to close, but thanks for this PR and thanks again for all your hard work on standardized-audio-context. been very helpful in this latest version of Tone.js!

PR #561, opened 2019-10-28
adding PitchShift example

chrisguttandin:
This pull request adds an example for the PitchShift effect. I initially created it to answer a [StackOverflow question](https://stackoverflow.com/questions/58574052/pitch-shifter-using-web-audio-api) and thought it might be worth to add it to the official docs.

I tried to follow the style of the other examples. Please let me know if there is anything I need to change.


tambien:
thank you for this PR!! very much appreciated

PR #567, opened 2019-11-05
Add `draw`, `now()`, `immediate()` to DummyContext

JackCA:
- this creates a new abstract BaseContext that DummyContext and Context
inherit from
- All methods available to Context are now stubbed by DummyContext
- BaseContext is now used as the type where Context was originally used

Note: `npm run test:node` still fails at `window undefined` but node should be able to use the esm build.

### DIFF
--- a/Tone/core/Global.ts
+++ b/Tone/core/Global.ts
@@ -1,29 +1,27 @@
 import { version } from "../version";
-import { hasAudioContext, theWindow } from "./context/AudioContext";
+import { AnyAudioContext, hasAudioContext, theWindow } from "./context/AudioContext";
 import { Context } from "./context/Context";
+import { DummyContext } from "./context/DummyContext";
+import { BaseContext } from "./context/BaseContext";
 import { OfflineContext } from "./context/OfflineContext";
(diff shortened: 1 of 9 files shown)


Tone/core/Global.ts
```diff
@@ -7,17 +7,35 @@ import { isAudioContext, isOfflineAudioContext } from "./util/AdvancedTypeCheck"
...
+	}
+	get destination() {
+		return {} as Destination;
+	}
+	get transport() {
+		return {} as Transport;
+	}
+}
+
+const dummyContext = new DummyContext();
```
tambien: the issue here is the Context class creates an AudioContext in the constructor, so this would be creating an AudioContext in this line which would still fail in a Node.js context. The more general solution would probably be to create an abstract class BaseContext file which DummyContext could extend.

JackCA: Good catch

JackCA:
@tambien this turned into a lot more churn, but I think it's a lot more robust now.

This seems to be the cleanest way to do this, although I wish there was a way to automatically implement the `BaseAudioContextSubset` methods in `BaseContext`.

Let me know what you think.

tambien:
Thanks for the helpful PR!

PR #576, opened 2019-11-13
Typescript: upgrade standardized-audio-context to 23.0.3

### LINKED ISSUE #574: Safari + Typescript Branch: Panner no sound
It seems that there's been a regression where a `Panner` node does not work in the typescript branch on safari.

I hear a small pop sound, but the same code works in `@next` 

[Example typescript](https://jsfiddle.net/JackCA/p1xLbq4n/8/)
[Example @next](https://jsfiddle.net/JackCA/gb3yq2ua/1/)

JackCA:
Fixes #574

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -806,19 +806,12 @@
       "dev": true
     },
     "automation-events": {
-      "version": "1.0.5",
-      "resolved": "https://registry.npmjs.org/automation-events/-/automation-events-1.0.5.tgz",
-      "integrity": "sha512-HOcyDR4gkKarfMwGvvCf7i/S87pLWV0JjG+j6d2DzfXNqUOjbSTZ/jliHNZqqYRoWxr+n9o0+zoC47ZzKFz4OA==",
+      "version": "1.0.6",
+      "resolved": "https://registry.npmjs.org/automation-events/-/automation-events-1.0.6.tgz",
+      "integrity": "sha512-uo7Cn4755zs26uPvFaAa4gUWv5IkBwiOoLH3Cc1KYS9A+NzmfbpN2AO2ycJ1/9V7yBd+4uaIoVewnQpO57TaWg==",
       "requires": {
-        "@babel/runtime": "^7.7.1",
+        "@babel/runtime": "^7.7.2",
         "tslib": "^1.10.0"
-      },
-      "dependencies": {
-        "tslib": {
-          "version": "1.10.0",
-          "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.10.0.tgz",
-          "integrity": "sha512-qOebF53frne81cf0S9B41ByenJ3/IuH8yJKngAX35CmiZySA0khhkovshKK+jGCaMnVomla7gVlIcc3EvKPbTQ=="
-        }
       }
     },
     "aws-sign2": {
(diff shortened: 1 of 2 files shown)


tambien:
i wonder what the FF issue is... will have to look into this

JackCA:
Shoot, didn't see that when running locally

JackCA:
I just gave this another go and thought I narrowed this down to it just being the Travis CI environment.

Attempted upgrading to the latest stable linux environment but it had no affect on the failing test.

The tests pass locally for me on firefox with a linux and mac, so the only way I can think to debug this is to be able to access the Travis CI debug mode (don't think that's possible for a public repo)

chrisguttandin:
Hi @JackCA, sorry for stepping in.

I tried to make sense of the stack trace. It somehow seems to be related to the new feature I added which allows to connect the destination to some other node. Do you have a chance to log some values on the Linux environment?

I'm curious what Firefox on Travis logs when running this code (without standardized-audio-context at all):

```JavaScript
const a = new AudioContext();
const o = new OfflineAudioContext({ length: 1, sampleRate: 44100 });

console.log(a.destination.channelCount, a.destination.channelCountMode, a.destination.channelInterpretation);
console.log(o.destination.channelCount, o.destination.channelCountMode, o.destination.channelInterpretation);
```

JackCA:
@chrisguttandin welcome :) 

Here's the output of that which I log out right before the `NotSupportedError`:

```
Firefox 70.0.0 (Linux 0.0.0) LOG: 0, 'explicit', 'speakers'
Firefox 70.0.0 (Linux 0.0.0) LOG: 1, 'explicit', 'speakers'
```

Any clues?

chrisguttandin:
For some reason the destination of the AudioContext has a channelCount of 0 which is not allowed. Maybe this is because there is not audio device available on the Travis machine.

JackCA:
Well literally the first thing I did when debugging this was try to replicate standardized-audio-context's travis settings -- I'm not sure if I mirrored it exactly but wouldn't you be running into the same issue in your tests?

chrisguttandin:
The standardized-audio-context tests use BrowserStack and SauceLabs. Maybe thinks are different with those services.

chrisguttandin:
It seems to be a bug in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1567949.

But the default is 2 in any case which is why we can set that directly without considering what the browser says.

I just applied a patch which should hopefully fix the issue: https://github.com/chrisguttandin/standardized-audio-context/commit/88c4fa7379661be86fc0375f28a14326bee0ac48.

JackCA:
@chrisguttandin awesome! Nice find. Really appreciate your help.

Will this warrant a version bump?

chrisguttandin:
Sure, I just published a patch release.

JackCA:
Jumped the version and there's good news and bad news: we don't have that error anymore, but we [have a new one](https://travis-ci.org/Tonejs/Tone.js/jobs/612498556?utm_medium=notification&utm_source=github_status) `IndexSizeError: Index or size is negative or greater than the allowed amount`

JackCA:
@chrisguttandin it looks like your integration tests are running against[ firefox 68](https://travis-ci.org/chrisguttandin/standardized-audio-context/jobs/612484403?utm_medium=notification&utm_source=github_status) whereas the latest version is 70.

Maybe that's why it's not getting caught?

chrisguttandin:
Hi Jack, the integration tests are only there to check that the isSupported promise resolves to false in unsupported browsers. Firefox v68 is the newest version of Firefox which is not supported anymore.

I just published another patch. Hopefully that solves the problem.

JackCA:
@chrisguttandin gotcha, that makes sense.

Unfortunately, we're still getting the [IndexSizeError](https://travis-ci.org/Tonejs/Tone.js/jobs/612612463) :/

chrisguttandin:
Sorry for the back and forth. It's a bit like poking in the dark. A made another attempt to fix the issue.

JackCA:
@chrisguttandin that was it --  we're all green now :champagne:

tambien:
@chrisguttandin @JackCA thank you so much for your efforts!

PR #586, opened 2019-12-06
Adding React+TypeScript cheatsheets to References

styk-tv:
There are tons of non-sense guides out there as to React+TypeScript. This one, I think, is worth linking to. And very up to date.


tambien:
Thanks for the PR, but doesn't seem relevant to include in the Tone.js since it deals entirely with React and Typescript. I've added the link to [this wiki](https://github.com/Tonejs/Tone.js/wiki/Using-Tone.js-with-React-or-Vue) since it seems more useful there.

Issue #622, opened 2020-02-27
github src vs. Tone.js src? And FR: add 'randomWalkNoRepeat' option


erichlof:
Hello!  First, thank you for this amazing library!  I was wanting to see how you implemented some of the pattern generators, 'randomWalk' in particular.  But I looked all over this repo's typescript source code and cannot find any mention of 'randomWalk' nor its implementation.  Yet, when I type 'randomWalk' in my project, it magically works!  So, it must be somewhere, lol.  I finally found it buried in the Tone.js download dist but it is minified so it is quite hard to read and understand.  

Am I missing something - why can't I find 'randomWalk'  anywhere on the Tone.js repo?  Is there another repo other than GitHub where you have newer features for the latest version of the library?

Also, a feature request:  I find myself wanting a combo of both randomOnce and randomWalk.  In other words, a melody would randomly walk from note to note, but would not repeat the same note from note to note.  This leads to more 'human' melody making on the piano for instance, A, B, C, B, C, D, E, F, E, D, etc. etc.  The current randomWalk comes close, but repeats, so I end up with A, B, B, B, C, B, C, C, D, D, D, D, C, etc.
btw I am a pro pianist/music teacher as well as a hobbyist coder (killer combo, huh? lol)

Maybe a 'randomWalkNoRepeat' would be useful to others as well?

Again, thanks for your awesome library!
-Erich

tambien:
Honestly i didn't know if anyone was using the CtrlPattern code on it's own, so during the refactor, i moved it to [PatternGenerator](https://github.com/Tonejs/Tone.js/blob/dev/Tone/event/PatternGenerator.ts) since it seemed like the primary use was within Tone.Pattern. 

Also seems like i may have removed the `randomWalk` pattern. If you're up for it and this is a useful feature for you, i'd happily review a PR for `randomWalk` and `randomWalkNoRepeat`.

erichlof:
@tambien 
Thank you for the response!  Yes I agree with the move to PatternGenerator - it makes most sense to live in the Tone.Pattern space.  The nice thing about randomWalk and the other CtrlPatterns is that you can let them operate on not just melody pitches, but other data that is inside a pattern array as well, like different rhythms, chords, etc..  The randomWalk in particular is great for linear melody making as it follows the human voice and other voice-like instruments which typically move smoothly from note-choice to note-choice, rather than randomly jumping to wide unrelated intervals.  As you may have guessed from my posts, I am investigating computer-generated music.
[...]

PR #625, opened 2020-03-02
Add 'randomWalk' pattern generator / add corresponding tests #622

erichlof:
See #622 
Implemented/improved 'randomWalk' as an additional choice for pattern generator.  The basic algorithm starts on a random chosen index inside the user-supplied pattern array.  It then randomly chooses whether to walk up or down 1 index in the array.  If it reaches the bottom of the array (index == 0), it is forced to walk up 1 step.  Likewise, if it reaches the top of the array (index == values.length - 1), it is forced to walk down 1 step.  This prevents the algo from getting stuck at the boundaries/edges of the array.  It will never repeat the same index (i.e. note) twice.

This pattern is desirable for users wanting to generate believable 'human-like' linear melodies because it smoothly steps between the values (i.e. note choices) much the same way as a vocalist or vocal-like instrument would randomly 'walk' through a melody or scale if asked to improvise.  

If a melodic leap is desired or needed in the randomly generated pattern, then users could alternate between 2 pattern generators:  this 'randomWalk' and 'random' (which might leap far from the current array index, or even possibly repeat the same note several times in row).  In this way, the two different pattern generators can work with each other.  For example, the currentIndex of the 'randomWalk' generator could be fed into the purely 'random' generator with a matching array list of the exact same possible note choices, and vice versa, thus producing more realistic, stylistic melodies.

### DIFF
--- a/Tone/event/PatternGenerator.test.ts
+++ b/Tone/event/PatternGenerator.test.ts
@@ -66,7 +66,7 @@ describe("PatternGenerator", () => {
 			expect(getArrayValues(pattern, 10)).to.deep.equal([4, 2, 3, 1, 2, 0, 1, 4, 2, 3]);
 		});
 
-		it("outputs random elements form the values", () => {
+		it("outputs random elements from the values", () => {
 			const values = [0, 1, 2, 3, 4];
 			const pattern = PatternGenerator(values, "random");
 			for (let i = 0; i < 10; i++) {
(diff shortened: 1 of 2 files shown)


erichlof:
@tambien 
Just checking in. Does everything look ok? Please let me know if I need to do anything else. 
Thanks! :-)

tambien:
@erichlof looks great! thanks for the PR and for the validation test

erichlof:
@tambien 
Thanks so much! I have been enjoying using your wonderful library, and will enjoy it even more with this new feature!
Thanks again,
-Erich

PR #631, opened 2020-03-17
Fixing bounds in TickSource Test

jsmith:
Just saw this small issue in one of the tests and thought I would make this PR! Pretty sure you meant `0.4`, right?


tambien:
great catch! thanks!

PR #636, opened 2020-04-05
Add dotted notes to the Subdivision type.

etherealmachine:
Reading https://github.com/Tonejs/Tone.js/blob/30e14cf7291fae180511c1ec7c998d4c752d4a52/Tone/core/type/Time.ts#L76 it looks like dotted notes can be returned, and in fact when I try it I can get a dotted note but the Typescript compiler complains that the return value cannot exist when used in a switch statement, so I think this is just an accidental omission.

2. Don't commit build files

### DIFF
--- a/Tone/core/type/Units.ts
+++ b/Tone/core/type/Units.ts
@@ -46,12 +46,13 @@ export type Positive = number;
 
 /**
  * Represents a subdivision of a measure.
- * The number represents the subdivision. "t" represents a triplet.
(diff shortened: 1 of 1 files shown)


etherealmachine:
I read through https://github.com/Tonejs/Tone.js/blob/dev/.github/CONTRIBUTING.md and didn't see anything on submitting PR's, apologies if this doesn't go to the correct place.

tambien:
🙏🏻thanks for the PR! overlooked the dotted notes when i migrated that to ts

PR #643, opened 2020-04-12
Use Triangle instead of Sawtooth for PulseOscillator source

Powerlated:
Using Sawtooth for the source wave produces an inaccurate tone at pulse widths other than 50% (0). The fix is to use a Triangle wave for the source.


tambien:
Great catch! not sure where i found the code for doing it as a sawtooth, but seems like triangle produces a much cleaner and more correct PulseOscillator

Fixed #643

PR #655, opened 2020-04-26
Changes to make Recorder work with polyfill

sanderwapstra:
I'm building a DAW for iPad where recording is important. The MediaRecorder API isn't available (yet) on Safari so a polyfill is needed. This seems to be the most used package for that: https://www.npmjs.com/package/audio-recorder-polyfill

`this._recorder.ondataavailable` is never fired in Safari (with the audio-recorder-polyfill) so I changed this to `this._recorder.addEventListener("dataavailable", handleData, false);`.

Another idea I had was to make `private _recorder: MediaRecorder;` public. Then you could use the MediaRecorder events, something like this `recorder.mediaRecorder.addEventListener('start', () => {}`.

What approach do you like best? Or do you have another idea?

Thanks!

I also tried https://www.npmjs.com/package/msr but couldn't get it to work. I’m getting a `Unhandled Rejection (Error): Recorder is already started` error after stopping the recorder on this line `Tone/component/channel/Recorder.ts:107`.

### DIFF
--- a/Tone/component/channel/Recorder.ts
+++ b/Tone/component/channel/Recorder.ts
@@ -102,16 +102,32 @@ export class Recorder extends ToneAudioNode<RecorderOptions> {
 		}
 	}
 
-	start(): this {
+	async start() {
 		assert(this.state !== "started", "Recorder is already started");
+		const startPromise = new Promise(done => {
+			const handleStart = () => {
+				this._recorder.removeEventListener("start", handleStart, false);
+
+				done();
+			};
+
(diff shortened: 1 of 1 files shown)


tambien:
This is really helpful! i think this approach is simpler than making the `_recorder` member public and then relying on the user to listen to the events. 

Also, if you suggest this works better with audio-recorder-polyfill, i'll also update the docs which link to the other one.

sanderwapstra:
Thank you for merging! I saw that you have already updated the documentation :)

PR #702, opened 2020-07-21
OneShotSource requestIdleCallback optimization

JackCA:
This optimization allows the browser to tell Tone a more opportune time to garbage collect during `onended` on `OneShotSource`. Otherwise, the 1 second timeout can trigger at an inopportune time.

It should be backwards compatible with [browsers that don't support `requestIdleCallback`](https://caniuse.com/#feat=requestidlecallback)

Note about formatting: I think a version of prettier may need to be defined because this is what's outputted when using the config in the `package.json` on my machine. The actual changes are [here](https://github.com/Tonejs/Tone.js/compare/dev...JackCA:oneshot-source-idle-callback?expand=1#diff-5a8d2ba7b2af136106e6c31eb70d53edR205-R212). I think it should be a dev dependency? And possibly a part of the linting command/s?

### DIFF
--- a/Tone/source/OneShotSource.ts
+++ b/Tone/source/OneShotSource.ts
@@ -1,5 +1,8 @@
 import { Gain } from "../core/context/Gain";
-import { ToneAudioNode, ToneAudioNodeOptions } from "../core/context/ToneAudioNode";
+import {
+	ToneAudioNode,
+	ToneAudioNodeOptions,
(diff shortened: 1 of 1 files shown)


JackCA:
Here's an example of the timers firing on a crowded frames:

Screen Shot 2020-07-20 at 8 20 20 PM

tambien:
awesome thanks!

PR #716, opened 2020-08-19
Allow the context to be resumed from an interrupted state.

louislepper:
As is the case when running on IOS, after coming back to the page from
the home screen.

This should fix a bug on IOS, where if you go to the home screen, or if your phone goes on standby, the audiocontext state changes to interrupted.

See similar fix: 
https://github.com/goldfire/howler.js/pull/1106
and https://github.com/goldfire/howler.js/pull/928

I wasn't able to get the tests running - npm install fails with ReferenceError: Unknown plugin "transform-runtime"
Even after manually running 
`npm install babel-plugin-transform-runtime`
The trail of missing dependencies seems to keep going for a while. Not sure what's happening there. I'm also on Node 11, as the travis build file specifies.

npm build seems to run fine.

Thanks!

### DIFF
--- a/Tone/core/context/Context.ts
+++ b/Tone/core/context/Context.ts
@@ -442,7 +442,9 @@ export class Context extends BaseContext {
 	 * to initially start the AudioContext. See [[Tone.start]]
 	 */
 	resume(): Promise<void> {
-		if (this._context.state === "suspended" && isAudioContext(this._context)) {
+		// @ts-ignore
+		if ((this._context.state === "suspended" || this._context.state === "interrupted")
+				&& isAudioContext(this._context)) {
 			return this._context.resume();
 		} else {
 			return Promise.resolve();


louislepper:
@tambien Would you be able to take a look at this for me?

Thanks!

chrisguttandin:
Hi @louislepper, hi @tambien, maybe this is something that should be handled by standardized-audio-context. What do you think that?

louislepper:
Potentially, although that'd involve changing this logic here, which resumes specifically based on the state.

I'm not actually sure why this code makes the if check here, is there any harm of calling resume on an audioContext that's not in a suspended/interrupted state?

tambien:
@louislepper thanks for the PR! @chrisguttandin it does seem to me like something that should be handled in standardized-audio-context since it's a browser-specific state and not something that's part of the Web Audio API spec.

louislepper:
@tambien any update?

tambien:
My preference would be that this is handled in standardized-audio-context since it's a browser-specific issue that might change over time. 

@chrisguttandin do you plan on including it in the lib? If Christoph _doesn't_ plan on including this in std-audio-context, then i'll be glad to merge this into Tone.js to fix this issue.

chrisguttandin:
Sorry for the delay. I'm definitely interested in fixing this with standardized-audio-context.

Since the goal of standardized-audio-context is to provide a spec compliant version of the AudioContext I would very much like to never have an AudioContext which says its state is `'interrupted'`. The `'interrupted'` state is not part of the spec and looks like something Apple just made up. Wether it's a good idea or not it's simply not spec compliant.

I build a little demo page to replicate the problem. From what I can tell it seems to be enough to keep an almost silent oscillator running in order to prevent Safari from setting the state to `'interrupted'`. @louislepper, could you please try if that works for you as well?

```js
const osc = Tone.context.createOscillator();
const gain = Tone.context.createGain();

gain.gain.value = 1e-37;
osc.connect(gain).connect(Tone.context.destination);
osc.start();
```

louislepper:
I had to use slightly different syntax - maybe we're on different versions:
`const osc = new Tone.Oscillator();
  const gain = new Tone.Gain(1e-37);
  osc.connect(gain)
  osc.start();
  gain.toDestination();`

But yep, that does seem to fix it.

Are you going to raise a PR to start a similar oscillator/sound automatically as part of the standardised audio context? Feel free to ping me when you do, otherwise I can try taking a look at it if I get time.

Thanks for the help!

chrisguttandin:
Hi @louislepper, thanks for verifying this. I did anyway work on a major update the weekend. So it was only a matter of adding another fix. This is the relevant commit: https://github.com/chrisguttandin/standardized-audio-context/commit/a6c967d276d8019a9cd7984191ecc768ab44455f.

I just released it as v25. It is a major version update because I also removed support for the old Edge v18 and below.

tambien:
Thanks again for the PR, went with #768 since it was slightly simpler and didn't include any iOS specific logic.

PR #722, opened 2020-08-26
fix: Instrument sync

Astray-git:
Fix some Instrument class overrides `sync` method without changing `this._sync` flag.


tambien:
good catch! very helpful thanks for the PR

PR #728, opened 2020-08-29
Docs fix

marcelblum:
just a 1 char docs bugfix :)
The topmost example code in the ToneAudioBuffers docs had a mistake and would not run.
Been using Tone a shitload lately and hope to contribute more.


marcelblum:
oops looks like I screwed up and merged master with dev before my commit? sorry...!

tambien:
good catch! could you please start from a clean PR so it doesn't include the merge commits?

PR #730, opened 2020-09-02
docs example bugfixes, typo corrections, minor additions for clarity,…

marcelblum:
here's my new & improved docs edits PR. Couple of example code bugs fixed, a few minor additions where I would have liked to have seen them when I was reading the docs, few typos fixed.

Couple docs-related questions for possible future PRs:

1. I noticed that a significant amount of classes are completely omitted from the Tone 14.x docs, apparently because their source contains no `@category` definition. E.g. Limiter, Recorder, MidSideCompressor, FrequencyEnvelope just to name a few. Was this intentional or just an oversight? AFAICT these classes seem to still be perfectly operational in the latest Tone.

2. One potential future project that would be really helpful for me and I'd guess other Tone users would be, in the Index section of the docs pages for each class, to automate creating separate sections for unique properties & methods and inherited ones. I think that could really help separate out the less useful descriptions of the generic shared properties/methods from the ones unique to that class and make them easier to read through. Also some basic navigational sugar such as a 'back to top' and ensuring the left hand column jumps to the selected anchor would be nice-to-haves. If you agree, any pointer you could offer into exactly where in the build process the docs are converted from the typedoc-generated json into the html/css seen at tonejs.github.io would be helpful.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -10,7 +10,7 @@ To contribute examples, please follow the current style of the examples. Add you
 
 There is always more work that can be done on documentation. Especially adding good examples to methods and members to make the docs more informative and useful for people coming from diverse musical and technical backgrounds.
 
-All of the docs are written in [jsdoc](http://usejsdoc.org/)-style comments in the source code. If you catch a mistake, please send a pull request.
+All of the docs are written in [TypeDoc](https://typedoc.org/)-style comments in the source code. If you catch a mistake, please send a pull request.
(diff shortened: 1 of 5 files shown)


tambien:
Thanks for the PR!

1) Yes those category omissions are just an oversight. Would be helpful to put those into the right category wherever you notice it. 
2) Modifications to how the doc pages get rendered would be [here](https://github.com/Tonejs/tonejs.github.io/tree/master/doc-gen)

PR #732, opened 2020-09-02
adding @category definitions for docs, fixing some typos/mistakes alo…

marcelblum:
Bunch of `@category` definitions enabling omitted/unbacklinked classes to appear in the docs index, plus a couple corrections.

I think I got every uncategorized class that was in the old Tone 13.x docs but missing from the v14 docs. Left some .ts files uncategorized because they seemed less important/internal helper classes and the like. I think if we add `"Global"` to the array at https://github.com/Tonejs/Tone.js/blob/5279a9d3d59dd329b89086c02a7400a8918b5d9b/scripts/typedoc.json#L6 then all uncategorized items would appear under a Global category in the docs index but that may be overkill.

### DIFF
--- a/Tone/component/channel/MidSideMerge.ts
+++ b/Tone/component/channel/MidSideMerge.ts
@@ -9,11 +9,12 @@ import { optionsFromArguments } from "../../core/util/Defaults";
 export type MidSideMergeOptions = ToneAudioNodeOptions;
 
 /**
- * MidSideMerge merges the mid and side signal after they've been separated by [[MidSideMerge]]
+ * MidSideMerge merges the mid and side signal after they've been separated by [[MidSideSplit]]
  * ```
  * Mid = (Left+Right)/sqrt(2);   // obtain mid-signal from left and right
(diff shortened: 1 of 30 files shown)


tambien:
Certain internal classes i never surfaced to the docs because those files aren't meant to be used by the end-user. a few examples are "ModulationSynth" which is the parent class to AM and FMSynth, but can't be used on its own. Definitely any class which is marked "abstract" will only clutter up the docs. The typedocs which appear on those classes are still surfaced the to the user on the child classes.  

Same with "Effect", "FeedbackEffect", "StereoEffect", etc. I would also add the "Tick" classes like "TickSource", "TickSignal", "Ticker" and the "Timeline" classes to that list of things which are internal and not exported so they can't be used externally. 

I can go through and mark each individually if it's unclear which ones don't need to be in the docs.

marcelblum:
Yes that makes total sense. I was mostly going by which classes already appeared at the top level of the old Tone 13.x docs e.g. https://tonejs.github.io/docs/13.8.25/Effect. I'll tidy this up and re-un-catergorize those items.

tambien:
Thanks for the help!

PR #735, opened 2020-09-07
Fixes #717

### LINKED ISSUE #717: Emitter callbacks with off removed incorrectly
I noticed a small bug with Emitter off implementation. It is using forward traversing and inline splice modification of the same array it is traversing. This causes element skips and if the same callback is attached twice in a row, it skips check of the latter one. Fix is to traverse the array in reverse so the array is nudged on already traveled elements.

https://github.com/Tonejs/Tone.js/blob/ed0d3b08be2b95220fffe7cce7eac32a5b77580e/Tone/core/util/Emitter.ts#L75-L79

Normally you would just have attached the same callback only once, so the bug does not appear even though element check is skipped. And it does remove multiple instances of the same callback fine if they are not attached next to each other.

jounii:
Reverses the loop for Emitter off. Added a test for this particular case.

### DIFF
--- a/Tone/core/util/Emitter.test.ts
+++ b/Tone/core/util/Emitter.test.ts
@@ -29,6 +29,19 @@ describe("Emitter", () => {
 		emitter.dispose();
 	});
 
+	it("can unbind duplicate events", () => {
+		const emitter = new Emitter();
+		const callback = () => {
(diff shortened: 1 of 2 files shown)


tambien:
thanks!

PR #753, opened 2020-09-23
Update Ticker.ts

sarahBuisson:
fix small ts error

2. Don't commit build files


tambien:
good catch!

i think the other solution for this that i've seen is define `_timeout` as `ReturnType<typeof setTimeout>` like so:

```ts
private _timeout!: ReturnType<typeof setTimeout>;
```

Do you have a sense of if either is better?

sarahBuisson:
@talbien  I don't know which solution is best, and haven't the time to explore which is the better one. I let it to you

PR #762, opened 2020-10-15
fix for #761

### LINKED ISSUE #761: regexp in buffer loading process causes hang with bracket characters in url, also is not specific enough
This only affects some very edge cases but can be nasty. When passing a file url to the ToneAudioBuffer constructor, if the url is long and contains square bracket characters early in the string, the regexp at
https://github.com/Tonejs/Tone.js/blob/ed0d3b08be2b95220fffe7cce7eac32a5b77580e/Tone/core/context/ToneAudioBuffer.ts#L362
totally freaks out and hangs the browser for up to several minutes, probably due to a quirk in javascript's regex engine (out of curiosity I tested the same expression with PCRE and it works fine, see https://regexr.com/5e633). While brackets are usually percent-encoded in a url string, in practice it's not always the case (especially when inputting user-provided urls), and also when accessing local files it won't be the case such as in an Electron app. Observed on Windows and Mac, Chrome and Firefox.

Here is a fiddle demonstrating in an actual url https://jsfiddle.net/marcelblum/5mvaorg6/27/ which hangs in Chrome but Firefox times out with a "too much recursion" error.

You can also just run a quick console test with an example local filepath string (might want to take a little walk after hitting enter!):
`("C:\\a folder\\another folder\\[something in brackets]\\yet another folder\\track 1.mp3").match(/\[(.+\|?)+\]$/)`

Further, I don't think this expression matches what's intended in certain cases. Looks like it's trying to catch urls with multiple extensions in brackets like "https://example.com/myfile.[flac|mp3|ogg]" but if you have a url with brackets earlier in the string like "https://example.com/some[dir]/myfile.[flac|mp3|ogg]" it will produce an incorrect match.

I'll submit a pr with a more specific regexp that fixes the issue by ignoring brackets earlier in the string.

marcelblum:
Improving url-parsing regexp to avoid hangs in edge case filepaths when loading a buffer from a file that contains brackets in its path string, fixes #761

### DIFF
--- a/Tone/core/context/ToneAudioBuffer.ts
+++ b/Tone/core/context/ToneAudioBuffer.ts
@@ -356,7 +356,7 @@ export class ToneAudioBuffer extends Tone {
 	static async load(url: string): Promise<AudioBuffer> {
 
 		// test if the url contains multiple extensions
-		const matches = url.match(/\[(.+\|?)+\]$/);
+		const matches = url.match(/\[([^\]\[]+\|.+)\]$/);
 		if (matches) {
 			const extensions = matches[1].split("|");
 			let extension = extensions[0];


tambien:
thanks! would be great to add a validation test if you have a moment

tambien:
I realize that a validation test might be a little difficult since you'd need to setup a server or file path that has a bracket in it. I guess if it's working with the current tests and you're able to get it to work with your scenario, that's probably good enough.

marcelblum:
ok cool. I have 0 experience writing automated tests so not sure I could have helped there anyway, sorry!

PR #766, opened 2020-10-23
Add docs

cjwit:
I added some detail to the documentation based on recent discussions in the Google Groups forum. This adds the HTML usage option to the main README file and adds some description and links between the Meter and DCMeter docs.

Related: is there a way to view the docs locally before pushing? Running `nom run docs` generated the JSON file, but I don't want to make changes to `package.json` to change that.

Since this is my first time contributing, I ran tests and saw some failures, but they don't relate to this pull request so far as I can tell. If I did break the tests somehow, please let me know so that I can figure out how that works. I would like to be able to contribute more often, even if only to incorporate things into the docs.

### DIFF
--- a/Tone/component/analysis/DCMeter.ts
+++ b/Tone/component/analysis/DCMeter.ts
@@ -4,9 +4,7 @@ import { MeterBase, MeterBaseOptions } from "./MeterBase";
 export type DCMeterOptions = MeterBaseOptions;
 
 /**
- * DCMeter gets the raw value of the input signal's waveform amplitude
- * at the current time and displays it as a value between 0 and 1.
- * See also {@link Meter}.
+ * DCMeter gets the raw value of the input signal at the current time. See also {@link Meter}.
  *
  * @example
  * const meter = new Tone.DCMeter();
--- a/Tone/component/analysis/DCMeter.ts
+++ b/Tone/component/analysis/DCMeter.ts
@@ -4,7 +4,9 @@ import { MeterBase, MeterBaseOptions } from "./MeterBase";
 export type DCMeterOptions = MeterBaseOptions;
 
 /**
- * DCMeter gets the raw value of the input signal at the current time.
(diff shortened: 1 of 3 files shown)


Tone/component/analysis/DCMeter.ts
```diff
@@ -4,7 +4,9 @@ import { MeterBase, MeterBaseOptions } from "./MeterBase";
 export type DCMeterOptions = MeterBaseOptions;
 
 /**
- * DCMeter gets the raw value of the input signal at the current time.
+ * DCMeter gets the raw value of the input signal's waveform amplitude
+ * at the current time and displays it as a value between 0 and 1.
```
tambien, line 8: I think the DCMeter is not limited to 0-1, it should be able to accept any number.

cjwit: When I use `.getValue()`, I see output values of -1 to 1 (which is an error and needs changed in this). The doc edit references that output, not the input. Am I understanding this incorrectly?

cjwit: I'm looking at this basic demo that I set up the other day: https://codepen.io/cjwit/pen/PozzrxP

tambien: The audio output of PolySynth is in the audio range (-1 to 1), but if you connect something with a different value, you can see that the values are not limited to that range: 

```js
const meter = new Tone.DCMeter()
const signal = new Tone.Signal(4).connect(meter)
```

cjwit: This is obviously related to me learning the system. Since a Signal appears to represent a wide range of types of values, it isn't clear to me what your example represents. (As in, I'm not sure what the "4" is relating to.) Since the point isn't troubleshooting here, I removed that language and pushed the change to the PR.

tambien:
thanks @cjwit Sorry to sleep on this! helpful PR

Issue #767, opened 2020-10-23
Feature request: a better way to resume audio context when "interrupted"


bryanbraun:
Currently, `Context.resume()` is great at resuming the audio context when the state is `suspended` (for example, when [the browser's autoplay policy](https://github.com/Tonejs/Tone.js/wiki/Autoplay) prevents audio on initial page load).

However, the audio context can also be put into a state of `interrupted`. I just learned that Safari and mobile Safari does this when there's an incoming call, browser dialog, or headphone jack removal (and possibly other situations). It appears this state is non-standard since [I don't see it in the spec](https://webaudio.github.io/web-audio-api/#BaseAudioContext).

When the state is `interrupted`, Tone's `Tone.context.resume()` does NOT resume the context (see [the code here](https://github.com/Tonejs/Tone.js/blob/aed85b0f47467b03209a0f0c435af01b1dc6f027/Tone/core/context/Context.ts#L441), which only checks for a `suspended` state). I expected this to resume the context, because the browser's `AudioContext.resume()` DOES resume the context (see [the examples in these webkit tests](https://github.com/WebKit/webkit/blob/master/LayoutTests/webaudio/audiocontext-state-interrupted.html)).

**My workaround:** find the AudioContext inside the ToneJS wrapper and run the `Tone.context._context.resume()` function. I'm not sure if it's a good idea for me to call an underscore method directly...

**Example (which includes steps to reproduce):** https://jsfiddle.net/bryanbraun/ou23xakm/18

<hr>

<small>Note: the workaround works, so I don't think this is urgent. It's just confusing to debug because I thought that Tone's `Context.resume()` was the same as the browser's `AudioContext.resume()`. Some documentation could improve this, but I think the long-term fix is making Tone's `Context.resume()` resume other states besides just `suspended`.</small>

<small>Thanks for the amazing library! It's been incredibly helpful in building out https://musicboxfun.com.</small>

bryanbraun:
I put together [a PR](https://github.com/Tonejs/Tone.js/pull/768), since my head was already in this space and it seemed straightforward. Let me know what you think!

chrisguttandin:
Hi @bryanbraun, I thought this was fixed already as mentioned here #716.

I'm not sure if @tambien published that already, though. Could you maybe try using tone@next?

bryanbraun:
Oh! I'm not sure how I missed that PR. 🤦 

[...]

PR #768, opened 2020-10-23
Address an edge-case for "Interrupted" states in Safari.

bryanbraun:
See #767 for details.

We _could_ check for an `interrupted` state here instead, but I thought it made more sense to fix the `interrupted` edge-case by having Tone's `resume()` match the underlying `resume()` more closely. Please let me know if I'm overlooking anything here!


tambien:
fixes #767

PR #789, opened 2020-11-15
Fix docs examples. 

### LINKED ISSUE #783: Bug: Oscillator docs - example for `partials` prop
Example code do not generate any results

```javaScript
const osc = new Tone.Oscillator("F3").toDestination().start();
setInterval(() => {
	// generate 8 random partials
        // osc.partials = new Array(8).map(() => Math.random()); // empty array
	osc.partials = new Array(8).fill().map(() => Math.random()); // add .fill() method before .map()
}, 1000);
```

Razz21:
Closes #783


Razz21:
Shouldn't be more intuitive to automatically start chorus' LFO when effect has been initialized? Is running LFO very expensive?

tambien:
thanks for the PR!

The Chorus could be initialized on init, but we'd need to add a new API to listen for when the Context is running, otherwise it would throw a warning/error that the context has not been started when the effect is created.

Issue #788, opened 2020-11-14
Jest & ReferenceError: AudioBuffer is not defined


sarahBuisson:
I use your lib, when I run it , it's okay , when I run my test with jest, it's okay but 

but when I try to make my test run with jest on the IDE to debug , I have this error :

"ReferenceError: AudioBuffer is not defined"

Any idea what it is? Where this class "AudioBuffer' is suppose to be defined?

sarahBuisson:
solution here: https://github.com/Tonejs/Tone.js/compare/dev...sarahBuisson:patch-3

FractalHQ:
I also have this issue when trying to use Tone with Svelte and Rollup.  Would be nice to have this merged.

guilhermecomum:
I'm getting this error too.

```
import Clave from "./clave.js";
import agogo1 from "../samples/brazilian-percussion/agogo1.wav";
import * as Tone from "tone";

describe("Clave class", () => {
  let samplers;

  beforeEach(() => {
    samplers = new Tone.ToneAudioBuffers({
      urls: { agogo1 },
    });
  });

  it("should instance a new Clave", () => {
    const clave = new Clave("10101", 120, "agogo1", samplers.get("agogo1"));
    expect.instanceOf(Clave);
  });
});

```

PR #790, opened 2020-11-15
Update AdvancedTypeCheck.ts

sarahBuisson:
fix https://github.com/Tonejs/Tone.js/issues/788

It was 'working" before my update because there is another classe AudioBuffer define in a file of the typescript lib d.ts who do something else, and don't need import to be used ( because it's "inside" typescript)

2. Don't commit build files

### DIFF
--- a/Tone/core/util/AdvancedTypeCheck.ts
+++ b/Tone/core/util/AdvancedTypeCheck.ts
@@ -1,6 +1,7 @@
 import {
-	isAnyAudioContext, isAnyAudioNode,
-	isAnyAudioParam, isAnyOfflineAudioContext,
+	AudioBuffer, isAnyAudioContext,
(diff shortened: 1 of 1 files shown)


tambien:
thanks for the PR! i think there'd be an issue with merging in this fix which is that it'll fail the build system which checks for variables that are defined but never used. 

When you say it was "working" before, what do you mean? i would imagine it would require including lib.d.ts since it relies on the Web Audio API which is defined in there.

chrisguttandin:
I think this patch will actually change the behavior since the AudioBuffer exported by standardized-audio-context is not the same as the global AudioBuffer available in the browser.

tambien:
finally had a chance to take a better look at this. thanks for the PR @sarahBuisson!

PR #815, opened 2021-01-01
allow JSON.stringify(audioContext)

JackCA:
This allows one to call `JSON.stringify(Tone.context)` without getting a circular reference issue. This matches what happens on a native `AudioContext` instance.

Also contains prettier changes based on the config in `package.json` as well (may be helpful to run prettier on the entire project)


tambien:
Thanks for the PR! i moved toJSON implementation to the BaseContext so that it doesn't need to be implemented twice, does this conflict with anything that you can think of?

JackCA:
@tambien nope, that's the better move! Thanks!

PR #826, opened 2021-01-20
fix the wrong variable name 'event' in Emitter.ts

vibertthio:
Fix the typo, `event` -> `eventName`.


tambien:
Good catch! thanks for the PR!

PR #831, opened 2021-01-24
Add buttons to the issue reproduction templates

vibertthio:
The old jfiddle template is using outdated API, such as `toMaster`. Also, there are warnings of starting audio contexts without user interaction. I fixed these in the new issue reproduction template.


tambien:
Thank you!

Issue #834, opened 2021-01-29
Feature Request: fanIn(), fanOut(), Tone.Tuner, Tone.Equalizer, Tone.ChannelEQ, Tone.EnvelopeFollower


Hazmatron7:
Hey I have some other cool ideas I would like to suggest to add for this library:

**1.** ```fanIn()``` and ```fanOut()```:
       

- basically its the same as ```ToneAudioNode.prototype.fan()``` but ```fanIn()``` does the opposite.
- Example: 
```javascript
/*
 *
 * Fan In:
 * 
 *  Node ---------->
 *                  \
 *                   -> Node
 *                  /        
 *  Node --------->
 * 
 * 
 * Fan Out:
 * 
 *           --------> Node   
 *          / 
 * Node --->
 *          \
 *           --------> Node
 * 
 */
```

**2.** ```"sawtoothUp"``` and ```"sawtoothDown"``` for ```Tone.LFO```:

- sawtooth on it's own, by default it goes up. To make it go down it easy, just invert the waveform.

**3.** ```Tone.Tuner```:

- It's a meter there where it gets frequencies from the waveform and detect notes from it.
- It should have at least a param called ```frequency``` to set a **A4** target reference to detect notes by. Default value should be 440Hz.
- It should be possible to implement using ```Tone.Analyser``` (or ```Tone.FFT```) and then calculate the frequencies to detect  notes and use an event or something to retrieve the results when it's detected a note.
- This could be very useful, especially when using midi notes.

**4.** ```Tone.Equalizer```:

- This could be very useful to use. I'm not sure exactly what type to suggest, either a graphic or parametric or something. But I'll leave that up to you to choose what's best.
- This is definitely achievable using BiquadFilters and or OnePoleFitler in series etc.

**5.** ```Tone.EnvelopeFollower```:

- It similar to ```Tone.AutoWah``` method  where it uses ```Tone.Follower``` as input and ```Tone.Scale``` as a output but with an AREnvelope and with sensitivity control.
- This could be useful for modulating an audio param from an other audio signal via envelope control

**6.** ```Tone.ChannelEQ```:
- This effect is used in Ableton Live. Here's an Image of it: 
![image](https://user-images.githubusercontent.com/48817592/106243460-602b6480-6259-11eb-900d-9be7b425b826.png)
- It's an effect with a 3 band gain controls and mid frequency and a toggle-able  80Hz highpass filter. Think of it as a 3 band equalizer.
[...]

PR #856, opened 2021-03-17
Add fanIn()

garrettmichaelgeorge:
This is a pull request for `fanIn()` as discussed in #834. `fanIn()` is essentially the inverse of `fan()`, connecting multiple input nodes to a single output node. Since, however, `fanIn()` involves multiple input nodes, it doesn't make sense to call it as a method on a single `ToneAudioNode`. I implemented it instead as a standalone function, a la `connect()`.

Feedback welcome!

### DIFF
--- a/Tone/core/context/ToneAudioNode.test.ts
+++ b/Tone/core/context/ToneAudioNode.test.ts
@@ -219,13 +219,13 @@ describe("ToneAudioNode", () => {
 			}, false);
 		});
 
-		it("can fan in multiple nodes to a destination", async () => {
+		it("can fan in multiple nodes to a destination", () => {
 			return PassAudio(input => {
-				const input0 = new Gain();
-				const input1 = new Gain();
-				const input2 = new Gain();
-				const output = new Gain();
-				fanIn(output, input0, input1, input2);
+				const context = input.context;
+				const gain0 = context.createGain();
+				const gain1 = context.createGain();
+				const output = context.destination;
+				fanIn(gain0, gain1, input, output);
 			});
 		});
 		
@@ -332,12 +332,13 @@ describe("ToneAudioNode", () => {
 		});
 
 		it("can fan in multiple nodes to a destination", async () => {
-			const output = new Gain();
-			const input0 = new Gain();
-			const input1 = new Gain();
-			const input2 = new Gain();
-			fanIn(output, input0, input1, input2);
-			expect(output.numberOfInputs).to.equal(3);
+			await Offline(() => {
+				const output = new Gain();
+				const input0 = new Gain();
+				const input1 = new Gain();
+				const input2 = new Gain();
+				fanIn(input0, input1, input2, output);
+			});
 		});
 	});
 
--- a/Tone/core/context/ToneAudioNode.test.ts
+++ b/Tone/core/context/ToneAudioNode.test.ts
@@ -3,7 +3,7 @@ import { Merge } from "Tone/component";
 import { Split } from "Tone/component/channel/Split";
 import { Oscillator } from "Tone/source";
 import { Gain } from "./Gain";
-import { connect, disconnect } from "./ToneAudioNode";
+import { connect, disconnect, fanIn } from "./ToneAudioNode";
 import { PassAudio } from "test/helper/PassAudio";
 import { Offline } from "test/helper/Offline";
 
@@ -218,6 +218,16 @@ describe("ToneAudioNode", () => {
 				disconnect(input);
 			}, false);
 		});
+
+		it("can fan in multiple nodes to a destination", async () => {
+			return PassAudio(input => {
+				const input0 = new Gain();
+				const input1 = new Gain();
+				const input2 = new Gain();
+				const output = new Gain();
+				fanIn(output, input0, input1, input2);
+			});
+		});
 		
 		it("can connect one channel to another", () => {
 			return PassAudio(input => {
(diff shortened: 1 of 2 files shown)


Tone/core/context/ToneAudioNode.ts
```diff
@@ -376,3 +377,16 @@ export function disconnect(
...
+/**
+ * connect the output of this node to the rest of the nodes in parallel.
+ * @example
+ * const player = new Tone.Player("https://tonejs.github.io/audio/drum-samples/conga-rhythm.mp3");
+ * const player1 = new Tone.Player("https://tonejs.github.io/audio/drum-samples/conga-rhythm.mp3");
+ * const filter = new Tone.Filter("G5").toDestination();
+ * // connect nodes to a common destination
+ * fanIn(filter, player, player1);
+ */
+export function fanIn(dstNode: InputNode, ...srcNodes: OutputNode[]): void {
```
tambien, line 390: I think the API would make more sense if it was `fanIn(...srcNodes, dstNode)` so that it's consistent with the rest of the connect APIs which always have the sources first and the destination after

Tone/core/context/ToneAudioNode.test.ts
```diff
@@ -320,6 +330,15 @@ describe("ToneAudioNode", () => {
...
 			});
 		});
+
+		it("can fan in multiple nodes to a destination", async () => {
+			const output = new Gain();
+			const input0 = new Gain();
+			const input1 = new Gain();
+			const input2 = new Gain();
+			fanIn(output, input0, input1, input2);
+			expect(output.numberOfInputs).to.equal(3);
```
tambien, line 340: `numberOfInputs` refers to the number of input channels, not connections, so i wouldn't expect the number of inputs to equal 3, it should still equal 1 like it did before.

Tone/core/context/ToneAudioNode.ts
garrettmichaelgeorge: That was my first thought also. The problem is that the rest parameter cannot come at the beginning of the function declaration – only the end.

How about this:

```typescript
export function fanIn(...nodes: any[]): void {
  const dstNode = nodes.pop();
  nodes.forEach(node => connect(node, dstNode));
}

fanIn(mySynth1, mySynth2, myReverb) // connects all to myReverb
```

The function declaration is less clear (and less strict), but it allows for parameters to include multiple `OutputNode`s and one `InputNode`.

garrettmichaelgeorge: See latest commit.

Hazmatron7:
@tambien @garrettmichaelgeorge 
Just a thought of a idea and or theory. Could this be a good way setting the parameters for the ```fanIn()```?
```javascript

/*

  @param srcNodes = Object[] | ToneAudioNode[] | AudioNode[]
  @param destNode = ToneAudioNode | AudioNode | Param | AudioParam |
*/
 function fanIn(...srcNodes, destNode) {...}

/*
 * where the last argument is the destination node
 * for input nodes it can be either a:
 *  {
 *    node: ToneAudioNode | AudioNode,
 *    input?: number,
 *    output?: number,
 *  }
 * or just ToneAudioNode | AudioNode on it's own with input and output set to 0
*/

fanIn(
   {node: player1, input: 0, output: 0},
   {node: player2, input: 0, output: 0}
   player3, //input and output set to 0
   filter, //destination
);
```
With this method it can set the input node connection buy choice. Especially things like ```Tone.Merge``` or native nodes (with multiple inputs). This just a suggestion to help out.
 
but after reading this:
>@tambien  "numberOfInputs refers to the number of input channels, not connections, so i wouldn't expect the number of inputs to equal 3, it should still equal 1 like it did before."

I see you mean.
but any thoughts?

Tone/core/context/ToneAudioNode.ts
```diff
@@ -376,3 +376,18 @@ export function disconnect(
...
+ * Connect the output of this node to the rest of the nodes in parallel.
+ * @param nodes One or more source nodes followed by one destination node
+ * @example
+ * const player = new Tone.Player("https://tonejs.github.io/audio/drum-samples/conga-rhythm.mp3");
+ * const player1 = new Tone.Player("https://tonejs.github.io/audio/drum-samples/conga-rhythm.mp3");
+ * const filter = new Tone.Filter("G5").toDestination();
+ * // connect nodes to a common destination
+ * fanIn(filter, player, player1);
+ */
+export function fanIn(...nodes: any[]): void {
```
tambien, line 390: better to not use `any[]` here. Maybe `OutputNode[]` (even though it's not totally correct)

tambien:
> ```js
> fanIn(
>    {node: player1, input: 0, output: 0},
>    {node: player2, input: 0, output: 0}
>    player3, //input and output set to 0
>    filter, //destination
> );
> ```

It's an interesting API, but to me feels just as verbose than how it would currently be written, so not sure what is afforded by the new API: 

```
player1.connect(filter, 0, 0);
player2.connect(filter, 0, 0);
player3.connect(filter);
```

Hazmatron7:
@tambien I understand. Keeping code writing simple is the a good method. Like I said it was just a suggestion to add to help out.  😄

Tone/core/context/ToneAudioNode.ts
garrettmichaelgeorge: Done – thanks for the feedback! See latest commit.

tambien:
thanks!

Issue #857, opened 2021-03-18
Calling dispose() on PolySynth after scheduled (using '+') triggerAttackRelease() creates error


kovatsgergo:
This works without error:
`xSynth= new Tone.MonoSynth().toDestination();
xSynth.triggerAttackRelease(330,1,'+1',1);
xSynth.dispose()`

This causes error:
`xSynth= new Tone.PolySynth().toDestination();
xSynth.triggerAttackRelease(330,1,'+1',1);
xSynth.dispose()`

Error message comes at around every 50 ms:
Uncaught Error: Synth was already disposed
    at r (Debug.ts:8)
    at pn._scheduleEvent (PolySynth.ts:241)
    at Object.callback (PolySynth.ts:253)
    at H._timeoutLoop (Context.ts:510)
    at H.emit (Emitter.ts:98)
r @ Debug.ts:8
_scheduleEvent @ PolySynth.ts:241
(anonymous) @ PolySynth.ts:253
_timeoutLoop @ Context.ts:510
emit @ Emitter.ts:98
15:13:54.250

I am also curious about how to kill all scheduled notes. Like a 'panic' or 'killall' function.
Using triggerAttackRelease() with '+' timing creates events seems to be unstoppable. At least not using Transport.stop() or releaseAll().

Any help is appreciated.

tambien:
Fixed by #859

PR #859, opened 2021-03-21
PolySynth does not reschedule event if disposed

stuartkeith:
I was trying to dispose of PolySynth instances during playback and had the `Synth was already disposed` assertion in `PolySynth._scheduleEvent` continually fail.

It turns out that `PolySynth._scheduleEvent` uses `setTimeout` to re-run `_scheduleEvent` if a note is occurring in the future, but that timeout does not account for the fact that the synth may be disposed in the meantime. So `_scheduleEvent` can be called again with `disposed === true`, and the assertion fails.

This PR simply checks `this.disposed` within the timeout to avoid the problem. I also added a test that fails without this new check.

This also seems to be the cause of issue #857.

### DIFF
--- a/Tone/instrument/PolySynth.test.ts
+++ b/Tone/instrument/PolySynth.test.ts
@@ -121,6 +121,20 @@ describe("PolySynth", () => {
 			});
 		});
 
+		it("can stop all sounds scheduled to start in the future when disposed", () => {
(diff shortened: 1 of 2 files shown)


tambien:
Super helpful! Thank you!

Fixes #857

PR #866, opened 2021-03-30
Fix typos and improve comments.

ronyeh:
Fixed a typo in `Monophonic.ts` with the word "immediately" and also simplified the description of the `velocity` parameter. 

Alternatively, the description could be rewritten as:
```
@param  velocity Determines how loud the note will be when triggered.
@param  velocity Determines the loudness of the note.
```


tambien:
thanks!

PR #884, opened 2021-04-25
Fix example for Part

matt-savvy:
The first example for Part doesn't play. 

## Changes
- adds start(0) to example code


tambien:
thank you!

PR #902, opened 2021-06-07
Fix: Loading samples with special characters in the file name

dannywritescode:
This PR enables loading samples with file names like `"C#5.wav"`. 

Currently, this throws an error because special characters like `"#"` need to be encoded in the right format before being passed into `fetch()` as the URL.

The solution is to use [`encodeURIComponent()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) to convert these characters into their UTF-8 encoded escape sequences (eg. `"#"` -> `"%23"`). This is done for each level of the file path to also encode any special characters that may be present in the folder names.

### DIFF
--- a/Tone/core/context/ToneAudioBuffer.ts
+++ b/Tone/core/context/ToneAudioBuffer.ts
@@ -372,7 +372,7 @@ export class ToneAudioBuffer extends Tone {
 
 		// make sure there is a slash between the baseUrl and the url
 		const baseUrl = ToneAudioBuffer.baseUrl === "" || ToneAudioBuffer.baseUrl.endsWith("/") ? ToneAudioBuffer.baseUrl : ToneAudioBuffer.baseUrl + "/";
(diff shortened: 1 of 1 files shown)


dannywritescode:
I found that the previous solution did not handle non-relative URLs properly. It would encode the colon in the protocol (eg `"https:"` -> `""https%3A"`), which threw errors when loading external samples as well as when loading data URIs. 

I have fixed this by creating a [`Location`](https://developer.mozilla.org/en-US/docs/Web/API/Location#properties) object through the `<a>` tag. This separates the domain, the protocol and the file path.

Now, only the file path is passed into `encodeURIComponent()`. And since `location.hash` stores the substring of the URL containing the `#` symbol and everything after it, the full file path becomes `location.pathname + location.hash`. 

I have tested this with external URLs, relative URLs, and a small wav data URI. It works as expected in all three cases! :)

PR #914, opened 2021-07-13
Fix browser entry point

benmccann:
UMD builds should be listed in `browser`. `main` is traditionally meant for the server-side entry point (i.e. CJS)

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#main
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#browser

The library currently won't load in Vite / SvelteKit because of this

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
 	"name": "tone",
 	"version": "14.8.0",
 	"description": "A Web Audio framework for making interactive music in the browser.",
-	"main": "build/Tone.js",
+	"browser": "build/Tone.js",
 	"module": "build/esm/index.js",
 	"type": "module",
 	"types": "build/esm/index.d.ts",


benmccann:
I think we could probably make `main` point to the `esm` build. Having `main` point the the UMD build which is only meant to be used in the browser is what's causing the problem.

I don't think bundlers can generally take UMD as input though I could be wrong on that. They definitely can take ESM though

benmccann:
@tambien did you have any thoughts about this?

tambien:
Do you know what the entry point that unpkg uses? There's a lot of people using this link: https://unpkg.com/tone, and i worry that if it all of a sudden pointed to the esm module instead of build/Tone.js, it might mess up some live sites.

benmccann:
Thanks for taking a look. I've updated it so that it will continue to work for unpkg.

From https://unpkg.com:
> If you omit the file path (i.e. use a “bare” URL), unpkg will serve the file specified by the unpkg field in package.json, or fall back to main.

benmccann:
Thanks for the merge! I have this same change pending in `tambien/Piano` if you'd be able to take a look and merge it there as well. Thanks!!

PR #940, opened 2021-09-01
Analyser constructor smoothing option bugfix

marcelblum:
Fix for the following bug: `smoothing` setting in the `Analyser` constructor is ignored (`smoothing` always ends up with the default value 0.8). Fixed by explicitly calling the `smoothingTimeConstant` setter in the constructor.


marcelblum:
Oh and just to clarify, to repro the bug:
```
const analyser = new Tone.Analyser({ type: "waveform", smoothing: .4, size: 512 });
console.log(analyser.smoothing);
//0.8
console.log(analyser._analysers[0]._nativeAnalyserNode.smoothingTimeConstant);
//0.8
```

PR #942, opened 2021-09-06
fix for AudioBufferSourceNode stop time miscalculation

### LINKED ISSUE #912: Player started with offsets intermittently plays erroneously past desired offset, depending on lookAhead
Been playing with taking a single buffer of a drum break and then using multiple `Player` objects to play different offset slices of different parts of the break, and noticed this bug during rapid fire drumpad triggering. Here is an example:

https://codepen.io/keymapper/pen/BaRLBMq

This creates 8 `Player` objects out of a short buffer and each plays a different slice of the drum break using offset parameters fed to `Player.start()` or `Player.restart()`. Each `Player` can be started by hitting a keyboard key Z through comma. The bug can be reproduced by rapidly hitting any one of those keys multiple times in succession. For example try smashing on the C key (snare sound) about a dozen times at a rate of at least 5 presses/second (you'll need at least 2 fingers :grin:) and you should hear the bug (e.g. you'll hear it play past the end of the snare sometimes). Observed in Chrome and Firefox on Windows 10.

Some notes:

- If lookAhead is set to 0, **the bug does not occur**, here is a version of that codepen with lookAhead at 0: https://codepen.io/keymapper/pen/dyWXEgR
- This is not an issue with `standardized-audio-context` as the bug occurs even when Tone is using a native audio context, here is a pen with a native audio context instead of Tone's default: https://codepen.io/keymapper/pen/OJmRJve?editors=0010 (note: this pen seems to only work on Chrome, won't work on FF, not sure if I am going about using a non-SAC context in Tone the wrong way here)

marcelblum:
Fixes #912 by using `now()` in place of `currentTime` when calculating the stop time for an audio buffer player, to allow for edge cases with `lookAhead` and end offsets set.

### DIFF
--- a/Tone/source/OneShotSource.ts
+++ b/Tone/source/OneShotSource.ts
@@ -168,7 +168,8 @@ export abstract class OneShotSource<
 
 		// schedule the stop callback
 		this._stopTime = this.toSeconds(time) + fadeOutTime;
-		this._stopTime = Math.max(this._stopTime, this.context.currentTime);
+		const nowTime = this.now();
+		this._stopTime = Math.max(this._stopTime, nowTime);
 		if (fadeOutTime > 0) {
 			// start the fade out curve at the given time
 			if (this._curve === "linear") {
@@ -188,7 +189,7 @@ export abstract class OneShotSource<
 				this._curve === "exponential" ? fadeOutTime * 2 : 0;
 			this._stopSource(this.now() + additionalTail);
 			this._onended();
-		}, this._stopTime - this.context.currentTime);
+		}, this._stopTime - nowTime);
 		return this;
 	}
 


marcelblum:
Ok sorry I will take a closer look when I get a chance. I had built Tone locally and tested manually before committing and everything looked/sounded good but perhaps there were unintended side effects. Will post here if I have any trouble getting the test suite running locally. Hoping [this wiki from 2018](https://github.com/Tonejs/Tone.js/wiki/Testing) is still current :)

tambien:
@marcelblum I spend some time last week getting karma to work on Node 14+. Have you pulled the latest? 

The tests that failed are: 

* ToneConstantSource Scheduling clamps stop time to the currentTime
* ToneOscillatorNode Scheduling clamps stop time to the currentTime

Pretty minor, might even just need a small tweak to those tests to account for the updated logic of the PR behavior.

PR #968, opened 2021-10-23
fix attempt #2 for AudioBufferSourceNode stop time miscalculation, with updated tests

marcelblum:
2nd try at a fix for #912, with tests updated to use the corrected time reference. All tests now pass on Mac (though not on Windows but that's a separate can of worms that has nothing to do with the changes made in this pr, see comments at #942).

I still plan to come up with a test to repro the conditions that lead to #912, will commit that when I get a chance.

### DIFF
--- a/Tone/signal/ToneConstantSource.test.ts
+++ b/Tone/signal/ToneConstantSource.test.ts
@@ -152,7 +152,7 @@ describe("ToneConstantSource", () => {
 				expect(source.getStateAtTime(0)).to.equal("stopped");
 				expect(source.getStateAtTime(currentTime)).to.equal("started");
 				setTimeout(() => {
-					currentTime = source.context.currentTime;
+					currentTime = source.now();
 					source.stop(0);
 					expect(source.getStateAtTime(currentTime + 0.01)).to.equal("stopped");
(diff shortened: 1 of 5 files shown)


marcelblum:
Hey @tambien here's a better fix for #912. Turns out the original theory of it being a `currentTime` vs `now()` issue was a red herring, though that did help me solve some other timing issues by accident.

Anyway for #912 I realized it was only happening on repeated calls to `player.restart()` and when there are multiple `player._activeSources`. `player.restart()` was leading to a call to `player._stop` which indiscriminately sets _all_ `_activeSources` to the same stop time. When there were multiple `_activeSources` with different delayed stop times due to `lookAhead` - such as when `restart()` is called rapidly at increments < `lookAhead` - this caused a pileup of increasingly erroneously delayed stops. My fix is to only stop the most recently created `source` on `player.restart()`, rather than stop all `_activeSources` (let me know if you think there might be a more elegant way to go about this).

marcelblum:
I've added a test for this fix (fails on dev, passes on this branch)

tambien:
Thank you! Really appreciate the contributions

PR #972, opened 2021-10-28
Parse note strings with three sharps or flats.

andypea:
Tweaked the note string regexp and the noteToScaleIndex map, so that it can parse note strings with three sharps or flats.

Although rare, notes with three sharps or flats do occur, for example in the B# augmented seventh and Cb diminished seventh chords.


andypea:
Is there anything I can do to help get this merged in?

tambien:
Thanks for the PR! sorry for the delay in merging.

PR #987, opened 2021-12-01
fixes for Context options handling & micro timing bugs

marcelblum:
This is mostly to fix the `Context` constructor's mishandling of `updateInterval` but also along the way fixes some other minor `Context` and `Ticker` constructor bugs, removes some unreachable/overridden/legacy code in `_setLatencyHint()`, and ensures that `updateInterval` is never unintentionally set higher than `lookAhead` which can lead to subtle timing issues. Also added a check for the expected `updateInterval` value in the `Context` constructor tests.

At first glance this probably looks like a lot of major internal plumbing refactors but it's not really! 😉  There are no breaking changes, all defaults remain as before; this is really just to address some bugs with edge cases when the user wants to roll their own `Context` using non-default values.

List of issues this addresses:

1) bugfix: `updateInterval` option in `Context` constructor is ignored:
```
console.log(new Tone.Context({updateInterval: .02}).updateInterval);
// 0.05
```

2) bugfix: `updateInterval` gets set to an unintended value in certain cases where non-default `latencyHint` or `lookAhead` are used, often greater than `lookAhead`, leading to potential subtle timing issues:
```
console.log(new Tone.Context({lookAhead: .1, latencyHint:"playback"}).updateInterval);
// 0.25
```

3) Changing `lookAhead` after context creation can lead to similar subtle timing issues because `updateInterval` doesn't automatically update along with it. Fixed by creating a `lookAhead` setter which sets the `updateInterval` to `( lookAhead / 2 )` in the spirit of the legacy `_setLatencyHint()`. When `lookAhead` is `0`, `updateInterval` defaults to `.01` which, admittedly, is just a value that I arrived at for best x-browser operation after some subjective testing on a few different devices, and can still be forced to a different value when explicitly set. See https://codepen.io/keymapper/pen/KKXKvPV in Chrome for a quick demo. It uses a short loud bass drum which results in an audible pop if the sound is clipped by the smallest amount, and uses `setInterval` rather than Tone scheduling in order to simulate a realtime source play trigger. When `lookAhead = 0` and `updateInterval` is left at the default `.05`, you'll hear a pop every 5-10 hits from the stop time getting miscalculated by a microsecond. When `updateInterval` is adjusted to `.01` the miscalculation seems *mostly* gone. (Note: due to timing/latency quirks this doesn't fix the micro-timing issue in all cases on all browsers though! I will open a separate issue for further discussion on this.).

4) bugfix: Minimum allowable `updateInterval` is miscalculated in certain cases because it is based on an assumed fixed 44100 `sampleRate`. Most non-Apple platforms seem to default to 48000, and the user can roll their own context at a wide range different rates depending on browser. Fixed by passing the context's `sampleRate` in with the `Ticker` constructor and calculating a lowest practical `_minimumUpdateInterval`. Also added the ability to pass an `updateInterval` of `0` to make it automatically set to the `_minimumUpdateInterval`.

5) bugfix: `Context.latencyHint` returns incorrect value in some cases when using a native `AudioContext`. This is mostly just an informational error, but this incorrect value was then being used by `_setLatencyHint()` to derive `updateInterval`. Since it's impossible for Tone to infer the `latencyHint` from a user-instantiated native context, I've changed `Context.latencyHint` to return an empty string when an own-rolled native context is passed in, and also set `updateInterval` to be calculated from `lookAhead` by default.
```
console.log(new Tone.Context({context: new AudioContext({latencyHint : "balanced"})}).latencyHint);
// "interactive"
```

6) Remove vestigial legacy handling of `latencyHint` in `_setLatencyHint()` much of which was unreachable code or getting overridden depending on constructor options. I think `_setLatencyHint()` was a holdover from older Tone versions where `latencyHint` in Tone had a different handling than the browser native `AudioContext.latencyHint`, perhaps from back when `latencyHint` was ignored by some browsers.

### DIFF
--- a/Tone/core/clock/Ticker.ts
+++ b/Tone/core/clock/Ticker.ts
@@ -16,7 +16,12 @@ export class Ticker {
 	/**
 	 * The update interval of the worker
 	 */
-	private _updateInterval: Seconds;
+	private _updateInterval!: Seconds;
+
+	/**
+	 * The lowest allowable interval, preferably calculated from context sampleRate
+	 */
+	private _minimumUpdateInterval: Seconds;
 
 	/**
 	 * The callback to invoke at regular intervals
@@ -33,11 +38,12 @@ export class Ticker {
 	 */
 	private _worker!: Worker;
 
-	constructor(callback: () => void, type: TickerClockSource, updateInterval: Seconds) {
+	constructor(callback: () => void, type: TickerClockSource, updateInterval: Seconds, contextSampleRate?: number) {
 
 		this._callback = callback;
 		this._type = type;
-		this._updateInterval = updateInterval;
+		this._minimumUpdateInterval = Math.max( 128/(contextSampleRate || 44100), .001 );
+		this.updateInterval = updateInterval;
 
 		// create the clock source for the first time
 		this._createClock();
@@ -121,9 +127,9 @@ export class Ticker {
 		return this._updateInterval;
 	}
 	set updateInterval(interval: Seconds) {
-		this._updateInterval = Math.max(interval, 128 / 44100);
+		this._updateInterval = Math.max(interval, this._minimumUpdateInterval);
 		if (this._type === "worker") {
-			this._worker.postMessage(Math.max(interval * 1000, 1));
+			this._worker?.postMessage(this._updateInterval * 1000);
 		}
 	}
 
--- a/Tone/core/context/Context.ts
+++ b/Tone/core/context/Context.ts
@@ -420,8 +420,8 @@ export class Context extends BaseContext {
 	}
 	set lookAhead(time: Seconds) {
 		this._lookAhead = time;
-		// if lookAhead is 0, default to .025 updateInterval
-		this.updateInterval = time ? (time / 2) : .025;
+		// if lookAhead is 0, default to .0125 updateInterval
+		this.updateInterval = time ? (time / 2) : .0125;
 	}	
 	private _lookAhead!: Seconds;
 
--- a/Tone/core/context/Context.ts
+++ b/Tone/core/context/Context.ts
@@ -3,7 +3,7 @@ import { Seconds } from "../type/Units";
 import { isAudioContext } from "../util/AdvancedTypeCheck";
 import { optionsFromArguments } from "../util/Defaults";
 import { Timeline } from "../util/Timeline";
-import { isDefined, isString } from "../util/TypeCheck";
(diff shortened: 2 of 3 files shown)


tambien:
Thanks for these updates!

Tone/core/context/Context.ts
```diff
@@ -116,26 +109,30 @@ export class Context extends BaseContext {
...
 		// fwd events from the context
 		this._context.onstatechange = () => {
 			this.emit("statechange", this.state);
 		};
-
-		this._setLatencyHint(options.latencyHint);
-		this.lookAhead = options.lookAhead;
+		
+		// if no custom updateInterval provided, updateInterval will be derived by lookAhead setter
+		this[arguments[0]?.hasOwnProperty("updateInterval") ? "_lookAhead" : "lookAhead"] = options.lookAhead;
```
tambien, line 135: I'm not fully following this line. Is there a reason not just to set `this.lookAhead = options.lookAhead`

Tone/core/clock/Ticker.ts
```diff
@@ -121,9 +127,9 @@ export class Ticker {
 		return this._updateInterval;
 	}
 	set updateInterval(interval: Seconds) {
-		this._updateInterval = Math.max(interval, 128 / 44100);
+		this._updateInterval = Math.max(interval, this._minimumUpdateInterval);
 		if (this._type === "worker") {
-			this._worker.postMessage(Math.max(interval * 1000, 1));
+			this._worker?.postMessage(this._updateInterval * 1000);
```
tambien, line 132: very minor, but the `!` in line 39 `private _worker!: Worker;` asserts that `_worker` is not null, so probably don't need this `?` null check here. Unless the compiler is complaining, i think this line could just be `this._worker.postMessage(this._updateInterval * 1000);`

Tone/core/context/Context.ts
marcelblum: Yes: The `lookAhead` setter I added also sets `updateInterval`. This line allows the dev to specify an `updateInterval` in the options if desired and not have it overridden by the `lookAhead` setter. In other words: If there is an `updateInterval` argument, set the private `_lookAhead` directly, otherwise use the `lookAhead` setter and allow that to also set `updateInterval`. This tries to cover all bases for example if the user wants a non-default `lookAhead` but didn't bother to specify any `updateInterval`.

tambien:
Thanks for this! 🙏🏻 Sorry took me a little while to pull it in.

PR #1004, opened 2021-12-28
Memoize getTicksAtTime and getSecondsAtTime

### LINKED ISSUE #370: repeated Transport pause() and start() causes degraded performance
My use case is very similar to the Transport Sync demo. I want to handle scenarios where the audio sees heavy replay, multiple seeks, etc. as users may want to edit the playback and fast-forward/rewind to listen to their changes. Over time, the performance slowly degrades after multiple Transport.pause() and Transport.start().

Easy repro:
```
let toggle = false;
setInterval(() => {
    toggle ? Transport.pause() : Transport.start();
    toggle = !toggle;
}, 50);
```
and let it run for a couple minutes.

On the Transport Sync demo `daw.html` after about ~1000 toggles the animation freezes running on a MacBook Pro (Retina, 13-inch, Early 2015). While this is a lot of playing and pausing, more complicated animations or audio tracks running on a lower powered device will freeze up a lot earlier.

This appears to be due to the internal state tracked by the underlying Clock/TickState/Timelines from the Transport on every pause and start. I'm still reading through the code here, but it doesn't seem necessary to track this state once play has resumed. Calling Transport.stop() clears all these internal events and restores performance, but also resets the timeline to the beginning, so this is only doable if we want to stop and reset all sources.

yifanmai:
Resolves #370

When TickSource.getTicksAtTime() or TickSource.getSecondsAtTime(), TickSource memoizes the ticks or seconds at the most recent event that has a state other than "started". The memoized values are used in subsequent calls.

### DIFF
--- a/Tone/core/clock/TickSource.ts
+++ b/Tone/core/clock/TickSource.ts
@@ -13,19 +13,19 @@ interface TickSourceOptions extends ToneWithContextOptions {
 	units: "bpm" | "hertz";
 }
 
-interface TickSourceOffsetEvent {
+interface TickSourceOffsetEvent extends TimelineEvent {
 	ticks: number;
 	time: number;
 	seconds: number;
 }
 
-interface TickSourceTicksAtTimeEvent {
+interface TickSourceTicksAtTimeEvent extends TimelineEvent {
 	state: PlaybackState;
 	time: number;
 	ticks: number;
 }
 
-interface TickSourceSecondsAtTimeEvent {
+interface TickSourceSecondsAtTimeEvent extends TimelineEvent {
 	state: PlaybackState;
 	time: number;
 	seconds: number;
@@ -112,6 +112,8 @@ export class TickSource<TypeName extends "bpm" | "hertz"> extends ToneWithContex
 			if (isDefined(offset)) {
 				this.setTicksAtTime(offset, computedTime);
 			}
+			this._ticksAtTime.cancel(computedTime);
+			this._secondsAtTime.cancel(computedTime);
 		}
 		return this;
 	}
@@ -133,6 +135,8 @@ export class TickSource<TypeName extends "bpm" | "hertz"> extends ToneWithContex
 		this._state.cancel(computedTime);
 		this._state.setStateAtTime("stopped", computedTime);
 		this.setTicksAtTime(0, computedTime);
+		this._ticksAtTime.cancel(computedTime);
+		this._secondsAtTime.cancel(computedTime);
 		return this;
 	}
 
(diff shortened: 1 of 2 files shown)


tambien:
Thank you for this PR @yifanmai 

It'd be great to add some tests which check edge cases like the assumption that the events are always increasing.

Tone/core/clock/TickSource.ts
```diff
@@ -3,7 +3,7 @@ import { Seconds, Ticks, Time } from "../type/Units";
 import { optionsFromArguments } from "../util/Defaults";
 import { readOnly } from "../util/Interface";
 import { PlaybackState, StateTimeline, StateTimelineEvent } from "../util/StateTimeline";
-import { Timeline } from "../util/Timeline";
+import { Timeline, TimelineEvent } from "../util/Timeline";
```
tambien, line 6: This line seems to be triggering a linting error which is causing Travis-CI to fail. I think `TimelineEvent` is not used, so removing it should fix that linting error.

Tone/core/clock/TickSource.ts
```diff
@@ -41,6 +53,16 @@ export class TickSource<TypeName extends "bpm" | "hertz"> extends ToneWithContex
 	 */
 	private _tickOffset: Timeline<TickSourceOffsetEvent> = new Timeline();
 
+	/**
+	 * Memoized values of getTicksAtTime at events with state other than "started"
+	 */
+	private _ticksAtTime: Timeline<TickSourceTicksAtTimeEvent> = new Timeline<TickSourceTicksAtTimeEvent>();
```
tambien, line 59: Thinking about that assumption which is now baked in that the events added to the timeline are always increasing, you can pass in an argument to Timeline which will assert that each time a value is added. 

```ts
new Timeline({
 increasing: true
})
```

tambien:
I think some tests would be extremely helpful. Especially tests which add events out of order and use `cancel`. That seems like the risk of memoization, that the memoized value might be out of sync with changes made to the timeline.

Tone/core/clock/TickSource.ts
yifanmai: Fixed. I'm now using the import to indicate that `TickSourceOffsetEvent`, `TickSourceTicksAtTimeEvent` and `TickSourceSecondsAtTimeEvent` are subclasses of `TimelineEvent`.

Tone/core/clock/TickSource.ts
yifanmai: I decided not to do add this assertion because it seems risky and might break things. Instead I changed the memoization to work with out-of-order events by invalidating all memoized values after the modified event.

yifanmai:
Thanks for the quick review! I've added tests for out-of-order events and cancel.

tambien:
Thank you for the PR!

PR #1005, opened 2022-01-05
Add Pattern.index

### LINKED ISSUE #934: Pattern.index is very needed
I'm building a flexible step sequencer based on the Tone.Pattern. I've got variable number of lanes with dynamic step count. 

I use empty arrays that get a `true` value when user activates a cell. Every lane has it's own note to play. Everything works fine, but I can't find a reliable way to get the current position of the pattern. 

The callback gets only `time` and `value` is either `undefined` or `true` and the `Tone.Pattern` in Tone v.14. doesn't have the `.index` feature, that was there in v.13. https://tonejs.github.io/docs/r13/Pattern#index

For now I'm using a simple `up` pattern and can set a separate `Tone.Pattern` to count and array of consecutive numbers, but I have issues with them getting out of sync. So it seems there is no solution to track the position of the Pattern in the `values` array. Especially when using the fancy random modes. 

Please, can we have the `Pattern.index` property back any time soon?

yifanmai:
Pattern.index returns the current index of the pattern. This method previously existed but was lost after the TypeScript conversion.

This changes PatternGenerator to return an index into the values array, rather than elements of the values array themselves. The index returned by PatternGenerator will be then returned by Pattern.index.

Fixes #934

### DIFF
--- a/Tone/event/Pattern.test.ts
+++ b/Tone/event/Pattern.test.ts
@@ -89,19 +89,21 @@ describe("Pattern", () => {
 		it("is invoked after it's started", () => {
 			let invoked = false;
 			return Offline(({ transport }) => {
+				const values = ["a", "b", "c"];
 				let index = 0;
 				const pattern = new Pattern((() => {
 					invoked = true;
-					expect(pattern.value).to.equal(index);
+					expect(pattern.value).to.equal(values[index]);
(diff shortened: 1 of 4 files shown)


tambien:
Thanks for adding all those tests! looks great

PR #1018, opened 2022-01-23
Smooth RMS values per channel in Meter

### LINKED ISSUE #882: Tone.Meter doesn't output stereo value correctly
Reproduce: 
- Go to [Mixer](https://tonejs.github.io/examples/mixer) example
- Solo Guitar 0
- Pan Guitar 0 all the way to left
- Play
- Observe that the right channel of the meter still pumps up, albeit lower than the left one.

Probably some changes underneath messed with the meter calculation?

yifanmai:
For multi-channel meters, smooth each channel independently using the channel's own saved last frame value, instead of using a shared last frame value that causes incorrect cross-channel smoothing.

Fixes #882


tambien:
amazing! thank you!

Issue #1037, opened 2022-02-16
New context is running mute


Azgaar:
**Describe the bug**
Trying to setup a new context with reduced sampleRate in `14.7.77`. Able to setup a context, but it's running mute.

````ts
  (Tone.getContext().rawContext as AudioContext).close(); // closing the current context
  const context = new AudioContext({ sampleRate: 24000 });
  Tone.setContext(new Tone.Context(context));
````

Found that I need to set the Destination after changing context. But how?
See https://groups.google.com/g/tonejs/c/WWURAv0n2wo/m/ehetsoU-BwAJ

Azgaar:
Yes, but I change the context on init step, before creating players/samplers.

Azgaar:
Here is an isolated part: https://stackblitz.com/edit/js-18cfrq?file=index.js

For some reason I'm getting a warning `'The AudioContext is "suspended". Invoke Tone.start() from a user action to start the audio.'` 
Invoking `Tone.start()` is not helping, and generally I don't see that new context is suspended.

marcelblum:
@Azgaar below is a quick simplified rewrite to get it working playing through the new context. Few things to note:

- use `Tone.getContext().dispose()` instead of `Tone.getContext().rawContext.close()` for a more thorough cleanup of Tone's default context
- Best not to do anything touching `Tone` until after `await Tone.start()` especially if creating a custom context, so I moved the calls to `setContext()` & `addPlayer()` & even `console.log()` to after that
- Keep in mind when you create a native context it lacks the advantages of Tone's bundled standardized-audio-context so will work less well across browsers and might be missing some functionality that Tone assumes it will have; can use `new Tone.context.rawContext.constructor()` instead of `new AudioContext()` to instantiate a standardized-audio-context, however 👇 
- 24000 is an odd sampleRate that some platforms might not support and standardized-audio-context will throw an error if the context's sampleRate !== the requested sampleRate
- The stuff you noted about `Transport._clock.context` seems like a bug, you can try creating new Transport using the new context as below but I haven't tested it much, can't promise it will work fully; much of Tone.js assumes you are using the default context so there are some subtle gotchas when using a custom context

```javascript
import * as Tone from 'tone';
import './style.css';

let player, context, transport;

const setContext = () => {
  Tone.getContext().dispose();
  const rawContext = new Tone.context.rawContext.constructor({
    latencyHint: 'playback',
    sampleRate: 24000
  });
[...]

PR #1041, opened 2022-02-22
actually close audio context when disposing it

marcelblum:
When looking into #1037 I found that, contrary to the docs, calling `context.dispose()` does not really close the audio context. Looks like just an oversight. This fixes the following bug:
```
Tone.getContext().dispose();
// ...
console.log(Tone.getContext().state); // "running", expected "closed"
```


marcelblum:
oops just noticed this causes some test fails, will look into soon

marcelblum:
this should do the trick and also ensure no existing code is broken by this change in case any users were already directly calling `context.close()` immediately after `context.dispose()` before the async `close` process resolves

tambien:
Thank you @marcelblum for this PR and for all your help in the repo in general! very much appreciated.

PR #1050, opened 2022-03-07
Use reciprocal of tempo when syncing time signals to Transport

### LINKED ISSUE #879: Delay time not reflecting transport bpm
My assumption is that delayTime of Tone.FeedbackDelay would be in sync with Tone.Transport.bpm .  Is this not the case?  The delayTime property itself works, but the timing is fixed and unrelated to transport bpm.

yifanmai:
Previously, signals with time units were synced to Transport by connecting it to the tempo signal, which has frequency units. This unit mismatch caused incorrect behavior, e.g. a time signal would double instead of halving as expected when bpm doubles. Now, signals with time units are connected to the reciprocal of the tempo signal instead, which also has time units.

fixes #879

### DIFF
--- a/Tone/core/clock/Transport.ts
+++ b/Tone/core/clock/Transport.ts
@@ -1,6 +1,8 @@
 import { TimeClass } from "../../core/type/Time";
 import { PlaybackState } from "../../core/util/StateTimeline";
 import { TimelineValue } from "../../core/util/TimelineValue";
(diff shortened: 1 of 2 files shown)


yifanmai:
On further thought, this PR has an issue whereby the new intermediate signal operators don't get garbage collected when the signal is unsynced. I'll fix this shortly.

tambien:
Great PR! thanks you!

Issue #1038, opened 2022-02-17
AudioWorklet error when using BitCrusher and PluckSynth


petethomasmusique:
First of all - many thanks again for the excellent Tone. I'm using it under the hood for parts of my PhD research and it's remarkable. 

**Describe the bug**
I get the following error when I instantiate a bitcrusher and karplus-strong synth:
<img width="664" alt="Screenshot 2022-02-17 at 20 07 17" src="https://user-images.githubusercontent.com/28138102/154562238-15a4cddf-93b3-42da-a97c-ec142206190e.png">

A description of what the bug is. 
The program still works, but I tend to get dropouts and/or I have to refresh numerous times before I hear sound again. FYI, I start Tone with user interaction, calling an asynchronous function. Then I start the Transport with a "+0.5" delay.

**To Reproduce**
```js
const crush = new Tone.BitCrusher()
const karplus = new Tone.PluckSynth()
```

chrisguttandin:
It looks like the `Context` is expecting that each processor get registered individually but currently all processors get registered together.

https://github.com/Tonejs/Tone.js/blob/b91b34d79293a7504f1dfc1552bfd4223dc2bb61/Tone/core/context/Context.ts#L361-L373

Do you agree @tambien? I could create a PR to update the `Context` to expect only one call to `addModule()` if you like.

tambien:
@chrisguttandin I'm not fully following, but if you have a fix for this, i would gladly accept the PR!

milianmori:
Hello everyone.

Experiencing almost the same issue.
I'm using Tone.Offline() to pre-render the audio to play it back later. If I pre-render it with Tone.BitCrusher() I'm getting this error when loading it on my website. (When loading on a localhost it works)

<img width="758" alt="Screenshot 2022-08-12 at 07 39 01" src="https://user-images.githubusercontent.com/100763596/184291590-7e474381-cf48-4586-8dd6-3216e934823f.png">


Cheers,
Milian

chrisguttandin:
Hi Milian, it looks like you're serving your site over http instead of https. Since an `AudioWorklet` runs on a high priority thread it can potentially get abused by bad actors. To mitigate the risk it can only be used on sites served over https. This rule doesn't apply when hosting a site on localhost to simplify the development process.

milianmori:
hmm okay thanks for the advice!

PR #1058, opened 2022-03-21
Load only a single AudioWorklet

chrisguttandin:
This PR changes the worklet loading code to expect only a single module and not a module per AudioWorklet. Before the code was expecting a module for each AudioWorklet but in fact only one module is used. This caused the bug reported here: #1038.

As always, please let me know if there is anything I should change.


tambien:
thanks!

cephasteom:
Hi @tambien and @chrisguttandin. Any progress on this being released into the main branch. Still seeing the issue. Many thanks.

Pete

PR #1094, opened 2022-06-29
docs typo fix

marcelblum:
Yes it's just one character, but it affects a not insignificant meaning change, and propagates into a lot of different Tone docs pages 😅


tambien:
haha! thank you!

PR #1113, opened 2022-09-03
Update README.md

djakinho:
Updates Synth instantiation.


tambien:
Thanks for the PR, these examples actually get executed on the tonejs homepage: https://tonejs.github.io/ so they currently need to have `Tone` as part of it.

PR #1120, opened 2022-09-27
Transport.progress: docs typo fix

satelllte:
Makes the description cleaner. 
Docs reference: https://tonejs.github.io/docs/14.7.77/Transport#progress

Thanks!

## CI issue

The issue https://github.com/Tonejs/Tone.js/issues/1123 still stays after 5 CI retries. I think it'll pass sooner or later eventually so I just marked this PR as ready for review


tambien:
Thanks @satelllte! One simple solution is to also change the codecov failure criteria to have a slightly higher threshold for failure.

PR #1131, opened 2022-10-15
allow worklet-based effects to be used with native contexts

marcelblum:
This is a fix for the following bug which occurs when trying to instantiate any effect node that uses an AudioWorklet (BitCrusher, JCReverb, etc.) if using Tone with a native context (non-Standardized Audio Context):

> Screenshot 2022-10-15 015036

If there's any better way than this to check if a context is from Standardized Audio Context or not, let me know! ~~This is using 
`context.constructor.name.includes("AudioContext")` which catches native online & offline contexts consistently AFAICT, whereas with those created by Standardized Audio Context `context.constructor.name` is always an empty string.~~ Now using `context instanceof window.BaseAudioContext` per @chrisguttandin's suggestion.


marcelblum:
Dang I wasn't seeing those tests fail locally, will investigate soon.

Tone/core/context/AudioContext.ts
```diff
@@ -47,9 +47,9 @@ export const hasAudioContext = theWindow &&
 	(theWindow.hasOwnProperty("AudioContext") || theWindow.hasOwnProperty("webkitAudioContext"));
 
 export function createAudioWorkletNode(context: AnyAudioContext, name: string, options?: Partial<AudioWorkletNodeOptions>): AudioWorkletNode {
-	assert(isDefined(stdAudioWorkletNode), "This node only works in a secure context (https or localhost)");
+	assert(isDefined(stdAudioWorkletNode), "AudioWorkletNode only works in a secure context (https or localhost)");
 	// @ts-ignore
-	return new stdAudioWorkletNode(context, name, options);
+	return new (context.constructor.name.includes("AudioContext") ? theWindow.AudioWorkletNode : stdAudioWorkletNode)(context, name, options);
```
chrisguttandin, line 52: Hi @marcelblum, maybe the following works a bit better.

```suggestion
	return new (context instanceof theWindow.BaseAudioContext) ? theWindow.AudioWorkletNode : stdAudioWorkletNode)(context, name, options);
```

marcelblum: Thanks @chrisguttandin, that's a lot more precise and seems to resolve the failing tests too

tambien:
🙏🏻 @marcelblum and @chrisguttandin

PR #1165, opened 2023-01-28
Use stricter types where possible

jdanford:


Tone/event/Sequence.ts
```diff
@@ -30,7 +30,7 @@ interface SequenceOptions<T> extends Omit<ToneEventOptions<T>, "value"> {
  * Tone.Transport.start();
  * @category Event
  */
-export class Sequence<ValueType = any> extends ToneEvent<ValueType> {
+export class Sequence<ValueType extends Time | null = Time | null> extends ToneEvent<ValueType> {
```
tambien, line 33: I'm not following the use of `Time` type here, if the sequence is like in some of the examples using Notes (e.g. `new Sequence((time, note) => {}, ["C4", "D4", "E4"])`), then the type should be `unknown | null` right?

tambien (changes requested):
Thank you for this helpful PR! i had just one question about the Sequence Generic Type, otherwise, it looks good to me!

Tone/event/Sequence.ts
jdanford: Ah yeah I was just looking at the tests and not considering the full context there – bounds have been removed!

jdanford:
@tambien this should be fixed now, can you take another look when you have a moment?

Issue #1051, opened 2022-03-12
Tone.js not playing audio on iOS


groudonzora:
I tried StackOverflow, but couldn’t get any responses. So I thought I’d try here.

I have a meteor application that uses Tone.js, and I use the GrainPlayer object specifically. It works perfectly fine on different browsers across different OS and devices… except for iOS. I personally don’t have an iOS device, but clients are showing me that iOS isn’t playing the audio. It’s also not restricted to safari; other browsers on their device also won’t play the audio.

I know it’s a Tone.js-specific issue here, because I also use Howler.js for a few of the sounds that don’t require the Tone GrainPlayer class, and those work. It’s also not an AudioContext issue in general, because I have the following code:

`Tone.setContext(Howler.ctx);`

Which ensures the AudioContext for Tone is routed through the AudioContext for Howler.

Is there a way to get it to play? I need to keep using GrainPlayer because of its unique properties and functions, so switching out will not work for me as a solution unfortunately. I have no way of diagnosing it from a developer standpoint because again I don’t have an iOS device, and even then they don’t have a way to pull up a developer console… But it seems to not break code or anything, just seems like it ‘ignores’ the sound playing, and moves on.

The sounds in question are created as such:

`standardChord[0] = new Tone.GrainPlayer(’/A.flac’).connect(Howler.masterGain);`

And then when they need to be played:

`standardChord[0].start();`

Which like I said works fine on other devices.

**More information:**
It's a Meteor application, bundled and deployed as a Node.js
The sounds do not start automatically; they start after a button press, and while it does use meteor's setInterval to play them, the Howler.js sounds do too, so that shouldn't be the issue either.

Willing to add code and/or additional details, if needed, not sure exactly what info might be useful (still new to Tone.js!)

dirkk0:
People will be able to help you debug this (I could and I will if you want to) but if you do professional client work, you need access to an iOS device because you will have these kinds of errors over and over again.

chrisguttandin:
I absolutely agree with @dirkk0. However there is one little detail which caught my attention in the few lines of code that you shared. It looks like you're trying to play a FLAC file. I doubt that this can be decoded on mobile Safari. I would be happy to hear that it works, though. You might want to try a different file format.

groudonzora:
> People will be able to help you debug this (I could and I will if you want to) but if you do professional client work, you need access to an iOS device because you will have these kinds of errors over and over again.

[...]

PR #1172, opened 2023-03-12
iOS solution via unmute.js

Utopiah:
cf https://github.com/Tonejs/Tone.js/issues/1051

2. Don't commit build files


tambien:
Thanks for the PR, but i dont' really recommend unmute.js anymore. i think better just to create a button and call `start` which is what's described in the README

PR #1219, opened 2023-11-27
Support for chords when using Sequence, use nested arrays to use chords

arjuna-dev:
2. Don't commit build files


tambien:
Thanks for the CL, but i think this would break the Sequence syntax. What i've done in the past for making sequences of chords was to use chord symbols (e.g. ["C", "Am", "Dm", "G"]) and then define a Map which maps the chord name to the array of notes

Issue #1026, opened 2022-02-03
There is no resume function for Tone.Recorder


samymsa:
The title says it all.

The MediaRecorder API actually does provide a [`MediaRecorder.resume()`](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/resume) function. But, unforunately, Tone.js doesn't.

This means that whenever I pause my Recorder using `Tone.Recorder.pause()`, I won't ever be able to resume it.
If I try to, I'll get the following error : `Failed to execute 'start' on 'MediaRecorder': The MediaRecorder's state is 'paused'`

tambien:
After calling `pause`, i believe you can restart the recording using `start()`.

samymsa:
That's what I thought, but doing this raises an error

Le ven. 4 févr. 2022 à 03:29, Yotam Mann ***@***.***> a
écrit :

PR #1266, opened 2024-08-02
Allow recorder to resume when paused

rupel190:
There is no straightforward way of resuming/unpausing a paused recorder as play() and pause() result in an error and the underlying MediaRecorder is exposed neither.

_Resumes recorder by calling start() as per expecation in issue #1026._

Tests are successful.


tambien:
Thank you! 

Fixes #1026

PR #1269, opened 2024-08-09
Fix link in the recorder documentation

matthewscharles:
The link to audio-recorder-polyfill on the Recorder documentation was broken -- just needed to swap the square brackets and parentheses.


tambien:
thank you!

PR #1293, opened 2024-11-28
Export remaining effect option interfaces

yannmazita:
The effect option interfaces of FeedbackDelay, FrequenceyShifter and Reverb (FeedbackDelayOptions, FrequencyShifterOptions and ReverbOptions) were not exported unlike all other effect option interfaces.
This PR fixes this discrepancy.


tambien:
Thanks!

PR #1328, opened 2025-03-29
fix(Filter): Use instance context in getFrequencyResponse

chufeng-official:
**Reason for Change:**  
In certain use cases, there is a need to use a newly created AudioContext instead of the default global context. However, the `getFrequencyResponse` method in the `Filter` class currently relies on the default global context. This occurs because the internally created `BiquadFilterNode` within `getFrequencyResponse` does not inherit the `Filter` instance's context.

**Solution:**  
This PR addresses the issue by explicitly passing the `Filter` instance's context to the constructor when creating the internal `BiquadFilterNode`. This ensures that frequency response calculations respect the `Filter`'s designated AudioContext rather than defaulting to the global context.


tambien:
thanks!

Issue #1138, opened 2022-11-02
Example of using audioWorklets (createAudioWorkletNode) with React


stevebarakat:
Can someone please create an example of using audioWorklets (createAudioWorkletNode) with React. Preferably and example of running Effects on audio worklet nodes.

marcelblum:
Not sure what would be different about using audioWorklet in React vs vanilla js. Maybe you can be more specific about what you're trying to do or code you've already tried?

stevebarakat:
That's fine. It doesn't have to be React. Is there a vanilla example anywhere? Using Tone's audioWorklet api, (not web audio api).

marcelblum:
Tone's AudioWorklet api is pretty basic, `createAudioWorkletNode()` just returns a `standardized-audio-context`-wrapped native `AudioWorkletNode` rather than a full-featured `ToneAudioNode`, meaning that it lacks Tone conveniences like `toDestination()` and requires a bit of careful handling to interact with other Tone components. This could use improvement in Tone.js, but I think the thinking behind it being implemented this way is because Tone can't infer much about the details of the custom processor being imported, so it's left to the user to handle specifics. Most of the heavy lifting is done inside the worklet js which has its own structure and syntax that is not Tone-specific (see https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Using_AudioWorklet).

Here's an example using the sample noise generator worklet from [here](https://googlechromelabs.github.io/web-audio-samples/audio-worklet/):
```js
await Tone.getContext().addAudioWorkletModule("https://cdn.jsdelivr.net/gh/GoogleChromeLabs/web-audio-samples/src/audio-worklet/basic/noise-generator/noise-generator.js");
const noiseWorkletNode = Tone.getContext().createAudioWorkletNode("noise-generator");
const gain = new Tone.Gain(2).toDestination();
noiseWorkletNode.connect(gain.input); // gain.input is necessary here because noiseWorkletNode is not a ToneAudioNode, so noiseWorkletNode.connect(gain) would fail

stevebarakat:
Thank you for the example. That was easy to implement with React but I'm still a bit confused... Most of the examples of audioWorklets are sound generators similar to that example but I'm interested in using audioWorklets with Tone.js components and effects like `Tone.Player`, `Tone.Meter`, `Tone.Delay`, etc–in order to improve the performance of my app (a multi-track audio mixer with fx busses). Is that possible?

marcelblum:
[...]

PR #1334, opened 2025-04-18
Support external AudioWorklets

### LINKED ISSUE #1326: What's the idea behind only allowing a single audio worklet module to be added?
This recent change for 15.x

> Load only a single AudioWorklet ([75a802c](https://github.com/Tonejs/Tone.js/commit/75a802c))

Is now only allowing to load a single audio worklet module. If you call `addAudioWorkletModule` two times with two different URLs, it will try to load the first worklet URL two times and cause a `NotSupportedError`. It will be obvious by looking at [this change](https://github.com/Tonejs/Tone.js/commit/75a802c#diff-7c0a0f73a79b99e24738a0c7627d1bdf84fc56e935745cf8de265becbb414e67).

chrisguttandin:
This PR is intended to fix https://github.com/Tonejs/Tone.js/issues/1326. It may also help with https://github.com/Tonejs/Tone.js/issues/1138.

The changes introduced in https://github.com/Tonejs/Tone.js/issues/1038 made it impossible to make Tone load any other `AudioWorklet` except for the ones build-in at least when using `addAudioWorkletModule()`.

To solve this the internal promise is now cached as a static variable on the `ToneAudioWorklet` class.

### DIFF
--- a/Tone/core/context/Context.ts
+++ b/Tone/core/context/Context.ts
@@ -346,9 +346,9 @@ export class Context extends BaseContext {
 	//--------------------------------------------
 
 	/**
-	 * Maps a module name to promise of the addModule method
+	 * A set of unsettled promises returned by the addModule method
 	 */
-	private _workletPromise: null | Promise<void> = null;
+	private _workletPromises = new Set<Promise<void>>();
 
 	/**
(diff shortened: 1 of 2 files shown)


tambien:
Thanks @chrisguttandin

Issue #1251, opened 2024-05-31
[Suggestion] upgrade dep eslint to v9


error-four-o-four:
Hey there,

typescript-eslint has released a beta [version](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8-beta) which supports the new eslint flat config format.

I'd be willing to create a draft or pr if you're interested.

And last but not least
thank you for your work 🤗

tambien:
Would be happy to review a PR if you wanted to contribute! i would very much appreciate it.

error-four-o-four:
Hey there,

I started working on the issue but I'd like to clearify a few things before finishing this pr

1. Is it desirable to enable [typed](https://typescript-eslint.io/getting-started/typed-linting) linting?
This would have a significant impact on performance but these linting rules could be used in a seperate script only.

2. The rule `@typescript-eslint/no-unused-expressions` keeps throwing an error
It looks like the rule was fixed in a previous [update](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v7.3.0) and I don't know what I should do. It doesn't seem correct to me to disable the rule.

3. `eslint-plugin-file-extension-in-import-ts` isn't included because it's not compatible with eslint v9 (yet)
One could use an [eslint-tool](https://github.com/eslint/eslintrc#readme) or use a customized rule like it's mentionend in the typescript-eslint [docs](https://typescript-eslint.io/troubleshooting/typed-linting/performance/#eslint-plugin-import) (see code snippet at the bottom of the page). I'd like to try the latter.

4. Honestly I do not have a lot of (or any) experience with webpack. I assume it's correct to exclude the files in `./examples/js`, right?

5. The most significant change I made was to seperate the scripts `lint` and `lint:all` to improve the performance. The former lints the source code in './Tone' (including the  test files) and the latter lints the whole project. I'm uncertain if it should be the other way round like `lint:src` and `lint` to lint the whole project by default.

Everything else should be self-evident (I guess).

---

FYI: [eslint-config-inspector](https://eslint.org/docs/latest/use/configure/debug#use-the-config-inspector) is a great tool to debug and inspect the eslint config

error-four-o-four:
Regarding the second question:

The rule throws here https://github.com/Tonejs/Tone.js/blob/dev/Tone%2Fcomponent%2Fanalysis%2FMeter.ts#L73-L74

tambien:
Thanks @error-four-o-four! This is hugely helpful! 

[...]

PR #1337, opened 2025-04-28
Upgrade eslint

error-four-o-four:
adresses #1251 

1. Please make all pull requests on the `dev` branch ✅
2. Don't commit build files ✅
2. Try to get all [tests](https://github.com/Tonejs/Tone.js/wiki/Testing) to pass ✅


test/scripts/test_integrations.mjs
```diff
@@ -7,7 +7,6 @@ const integrations = await glob(resolve(__dirname, "../integration/*"));
 for (let dir of integrations) {
 	await within(async () => {
 		cd(dir);
-		// eslint-disable-next-line no-console
```
tambien, line 10: I would like to keep on no-console as a rule if possible. Sometimes when i'm debugging i use a lot of console.logs and would love to catch it before it's submitted.

Issue #1339, opened 2025-05-05
generate_docs.cjs is missing


CalvinJamesHeath:
**Describe the bug**

Attempting to run npm run docs in the Tone.js repository results in an error:
`Error: Cannot find module '/root/Tone.js/scripts/generate_docs.cjs'`

**To Reproduce**

1. Clone the Tone.js repository.
2. Navigate to the project directory.
3. Run npm install to install dependencies.
4. Run npm run docs.

**Expected behavior**
The npm run docs command should execute the generate_docs.cjs script to generate the project's documentation.

**What I've tried**
1. Verified that the generate_docs.cjs file is indeed missing from the scripts directory.
2. Checked the repository's branches and commit history to see if the script was removed or relocated.
3. Searched for any documentation or issues referencing the generate_docs.cjs script.

tambien:
good catch, i think that this npm script is no longer used, probably can be removed. 

[The docs build](https://github.com/Tonejs/tonejs.github.io/blob/master/scripts/build_latest.mjs#L33) script actually only calls `npm run build:json`

PR #1348, opened 2025-05-27
investigate deps

error-four-o-four:
1. Please make all pull requests on the `dev` branch. ✅
2. Don't commit build files ✅
3. Try to get all [tests](https://github.com/Tonejs/Tone.js/wiki/Testing) to pass. ✅

---

Hey there,

as already mentioned I began to migrate dependencies  `fs-extra`, `glob`, `rimraf`

Additionally .mjs and .cjs files were included in the linter config. FYI: `simple-sort-import` does not work with commonJS files ([link](https://github.com/lydell/eslint-plugin-simple-import-sort?tab=readme-ov-file#does-it-support-require)). This could be a reason to migrate the scripts. I'd be happy to update this PR and implement this change if you're interested.

I didn't investigate other dependencies yet but [npmgraph](https://npmgraph.js.org/?q=tone%4015.2.6#deps=devDependencies&color=outdated) provides a good entrypoint to do so.

A few other things that caught my attention but aren't implemented yet:
* ~~The npm script `docs` fails because `scripts/generate_docs.cjs` does not exist~~ (has issue)
* I recently switched from windows to WSL/Ubuntu and had to explicitly set `executablePath` to make puppeteer happy and successfully run the tests (I think this has a very low priority)
* ~~Could you please elaborate why `tslib` is included as a dependency? I guess it's because the .ts files are also included via the field `files` in `package.json`, right?~~


test/scripts/test_html.cjs
```diff
@@ -46,7 +46,16 @@ async function testExampleString(str) {
...
+const htmlFiles = globSync(
+	"*.html",
+	{
+		cwd: resolve(__dirname, "../../examples"),
+		absolute: true
+	}
+);
+
+console.log(htmlFiles);
+process.exit();
```
error-four-o-four: fixed in 3b094ee

error-four-o-four:
closes #1339  

Oh dear.

After all this PR reduces the following stats:
* Amount of dependencies (including devDependencies): 927 => 904
* Modules with multiple versions: 80 => 78
* Deprecated Modules: 10 => 3

but there are still a few things to do:

[@rollup/plugin-commonjs](https://www.npmjs.com/package/@rollup/plugin-commonjs)
[@types/chai](https://www.npmjs.com/package/@types/chai)
[@types/mocha](https://www.npmjs.com/package/@types/mocha)
[codecov](https://www.npmjs.com/package/codecov)
[concurrently](https://www.npmjs.com/package/concurrently)
[http-server](https://www.npmjs.com/package/http-server)
[plotly.js-dist](https://www.npmjs.com/package/plotly.js-dist)
[semver](https://www.npmjs.com/package/semver)
[typedoc](https://www.npmjs.com/package/typedoc)
[ua-parser-js](https://www.npmjs.com/package/ua-parser-js)
[webpack-cli](https://www.npmjs.com/package/webpack-cli)
[yargs](https://www.npmjs.com/package/yargs)

When using `@rollup/plugin-commonjs` >=27 the tests fail because external modules can't be resolved. I've tried to add `"allowSyntheticDefaultImports": true` to the compiler options but without any success. I'd have to investigate the webpack bundle process and tools a bit further. At the moment I'm very tempted to migrate to `vite` and `vitest` but I don't know if this is a viable option (esp. in regard to the workflows of @tonejs/ui and tonejs.github.io) but not today.

package.json
```diff
@@ -15,20 +15,20 @@
 		"Tone"
 	],
 	"scripts": {
-		"build": "npm run increment && rimraf build && npm run ts:build && npm run webpack:build",
-		"docs": "node scripts/generate_docs.cjs",
-		"docs:json": "typedoc --options \"./scripts/typedoc.json\" --json \"./docs/tone.json\"",
+		"remove": "node -e \"fs.rmSync('./build', { recursive: true, force: true })\"",
+		"build": "npm run increment && npm run remove && npm run ts:build && npm run webpack:build",
+		"docs": "typedoc --options \"./scripts/typedoc.json\" --json \"./docs/tone.json\"",
```
tambien: could we keep this as docs:json, this script is actually referenced in the docs generating script [here](https://github.com/Tonejs/tonejs.github.io/blob/master/scripts/build_latest.mjs)

tambien (commented):
thanks for this!

package.json
```diff
@@ -15,20 +15,20 @@
 		"Tone"
 	],
 	"scripts": {
-		"build": "npm run increment && rimraf build && npm run ts:build && npm run webpack:build",
-		"docs": "node scripts/generate_docs.cjs",
-		"docs:json": "typedoc --options \"./scripts/typedoc.json\" --json \"./docs/tone.json\"",
+		"remove": "node -e \"fs.rmSync('./build', { recursive: true, force: true })\"",
```
tambien, line 18: also out of curiosity, why remove `rimraf`? seems like the code was more succinct before

package.json
error-four-o-four: My bad. I didn't know that it's used in the homepage. I'm wondering if it's worth to take a look at the typedoc API or if this should be adjusted in any kind of way?

package.json
error-four-o-four: `rimraf` pulls in many transitive dependencies. The builtin node module provides the same functionality. The Disadvantage is that it depends on the Nodejs runtime which excludes older Nodejs versions, Bun and Deno users. If it's desired to support them I'd recommend to use [premove](https://www.npmjs.com/package/premove) which has zero dependencies.

I'm uncertain if 'remove' is the best name for this script. One could also name it 'clean' or use it as the npm lifecycle script 'prebuild'.

Less dependencies means a smaller footprint and less maintenance cost

feels good

tambien:
thank you!! really appreciate the time you put into making the library simpler and more maintainable 🙏🏻

Issue #1310, opened 2025-01-20
Tone.Sampler loop in the same way than Tone.Player


josephernest:
Hi!
Does [Tone.Sampler](https://tonejs.github.io/docs/15.0.4/classes/Sampler.html) support loops (different loops for each different audio files) in the same way Tone.Player does:
```
const player = new Tone.Player({url: `test.wav`});
player.loopStart = 44100;
player.loopEnd = 88200;
player.loop = true;
player.start();
```
?

Thanks!

PR #1350, opened 2025-06-04
Adding loop functionality to Sampler (#1310)

Caden-Hornyak:
This PR adds looping functionality to the Sampler consistent with Tone.Player. Looping is useful for Sampler instruments that need permanent continuous sound through crossfading such as the Violin.

An earlier PR by maximelebreton's from 2023 introduced basic looping but it lacked Unit testing as well as consistent behavior with Tone.Player such as getter's/setter's.

### DIFF
--- a/Tone/instrument/Sampler.ts
+++ b/Tone/instrument/Sampler.ts
@@ -10,9 +10,9 @@ import {
 	Note,
 	Time,
 } from "../core/type/Units.js";
-import { assert } from "../core/util/Debug.js";
+import { assert, assertRange } from "../core/util/Debug.js";
 import { timeRange } from "../core/util/Decorator.js";
-import { optionsFromArguments } from "../core/util/Defaults.js";
+import { defaultArg, optionsFromArguments } from "../core/util/Defaults.js";
 import { noOp } from "../core/util/Interface.js";
 import { isArray, isNote, isNumber } from "../core/util/TypeCheck.js";
 import { Instrument, InstrumentOptions } from "../instrument/Instrument.js";
@@ -34,6 +34,9 @@ export interface SamplerOptions extends InstrumentOptions {
 	baseUrl: string;
 	curve: ToneBufferSourceCurve;
 	urls: SamplesMap;
+    loop: boolean;
+    loopEnd: number;
+    loopStart: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Tone/instrument/Sampler.ts
```diff
@@ -357,6 +393,90 @@ export class Sampler extends Instrument<SamplerOptions> {
...
+        this._providedMidiNotes.forEach((midiNote) => {
+            const buffer = this._buffers.get(midiNote);
+            if (buffer.loaded) {
+                assertRange(this.toSeconds(loopEnd), 0, buffer.duration);
+            }
+        });
+        // get the current sources
+        this._activeSources.forEach((sourceList) => {
+            sourceList.forEach((source) => {
+                source.loopStart = loopEnd;
```
tambien: Looks like you accidentally set loopStart instead of loopEnd here.

Tone/instrument/Sampler.ts
```diff
@@ -197,6 +226,7 @@ export class Sampler extends Instrument<SamplerOptions> {
 		if (!Array.isArray(notes)) {
 			notes = [notes];
 		}
+        const offset = defaultArg(0, this._loopStart);
```
tambien: I think you have this reversed, should be 

`const offset = defaultArg(this._loopStart, 0);`

Tone/instrument/Sampler.ts
```diff
@@ -210,16 +240,22 @@ export class Sampler extends Instrument<SamplerOptions> {
 			const playbackRate = intervalToFrequencyRatio(
 				difference + remainder
 			);
+            let duration = this._loop
```
tambien: nit: if duration isn't redefined, could be `const` instead of `let`

Tone/instrument/Sampler.ts
```diff
@@ -357,6 +393,90 @@ export class Sampler extends Instrument<SamplerOptions> {
 		return this._buffers.loaded;
 	}
 
+    /**
+     * Set the loop start and end. Will only loop if loop is set to true.
+     * @param loopStart The loop start time
+     * @param loopEnd The loop end time
+     * @example
+     * const sampler = new Tone.Sampler("https://tonejs.github.io/audio/berklee/guitar_chord4.mp3").toDestination();
```
tambien: nit: this `@example` isn't a valid use of the API. i think it'd need to look like: 

```ts
const sampler = new Tone.Sampler({
	urls: {
		A1: "https://tonejs.github.io/audio/berklee/guitar_chord4.mp3",
	},
}).toDestination();
// loop between the given points
sampler.setLoopPoints(0.2, 0.3);
sampler.loop = true;
```

tambien (commented):
thank you so much for adding this functionality, it is very often requested! 

a couple mostly very small nits in the PR and otherwise it looks good!

Caden-Hornyak:
I've pushed fixes for the nits and linting errors. Let me know if anything else needs adjusting!

tambien:
great work! thanks so much!

PR #1355, opened 2025-07-09
Support sampleRate option for realtime AudioContext

### LINKED ISSUE #1023: add sampleRate option to Tone.Context()
Is it possible to allow setting the sampleRate to the ```Tone.Context()``` as a constructor parameter on initialization please?

```javascript
    //example:
    let context = new Tone.Context({
        latencyHint: "interactive",
        sampleRate: 44100,
        //other options etc.....
    });
```

Basically it should be that if parameter has a valid number use it or if undefined then use automatic detection by default (which is built in already)

The only error it will throw is this:
image

which can be useful to detect that the sampleRate isn't supported with a ```try/catch```

And yes according to caniuse.com and MDN, It is supported to all browser now.
Except for IE of course but who cares about IE anymore lol 😆 

image
image

chrisguttandin:
This PR is intended to fix #1023. It's adding the possibility to set the `sampleRate` when creating a new realtime `AudioContext`.

Does the documentation update automatically?

Should `sampleRate` better be of type `Frequency` instead of being a simple `number`?

### DIFF
--- a/Tone/core/context/Context.test.ts
+++ b/Tone/core/context/Context.test.ts
@@ -74,11 +74,13 @@ describe("Context", () => {
 				latencyHint: "playback",
 				lookAhead: 0.2,
 				updateInterval: 0.1,
+				sampleRate: 32000,
 			});
 			expect(ctx.lookAhead).to.equal(0.2);
 			expect(ctx.updateInterval).to.equal(0.1);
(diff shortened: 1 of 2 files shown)


tambien:
thank you @chrisguttandin

Issue #1354, opened 2025-06-19
MembraneSynth documentation describes valid `octaves` range incorrectly


zerobias:
[Documentation](https://tonejs.github.io/docs/15.1.22/classes/MembraneSynth.html#octaves) and [source code](https://github.com/Tonejs/Tone.js/blob/dev/Tone/instrument/MembraneSynth.ts#L33) states that max value of `octaves` is 8, but even default value [is 10](https://github.com/Tonejs/Tone.js/blob/dev/Tone/instrument/MembraneSynth.ts#L76)

Even more, this parameter is not an octave number at all

```ts
const maxNote = hertz * this.octaves;
```

This operation yield 3.3 octaves at value 10

tambien:
good catch!

PR #1359, opened 2025-08-08
Issue #1354 fix - 'MembraneSynth'

aamcn:
Issue link:  (https://github.com/Tonejs/Tone.js/issues/1354#issue-3159829067)

Correct octave default value to match documentation.
Add range values for octaves and pitch decay timeRange to match documentation.
Update maxNote equation to increase in full octaves. 
Add update or add tests for the above changes.


tambien (commented):
really appreciate the fix! Could you remove the changes to package-lock and package? seems like they may have been made in error, thanks!

Issue #1365, opened 2025-08-30
[Feature Request] Reverse Delay


yosbelms:
**The feature you'd like**
Reverse Delay Effect

**Any alternatives you've considered**
Tried using GrainPlayer, did not achieved the desired effect.

**Additional context**
This doc contains some ideas to extract.
https://ccrma.stanford.edu/~jingjiez/portfolio/echoing-harmonics/pdfs/A%20Pitch%20Shifting%20Reverse%20Echo%20Audio%20Effect.pdf
More info: https://gearspace.com/board/showpost.php?p=15709467&postcount=922

PR #1395, opened 2025-12-26
feat: Reverse delay

jswill88:
Issue #1365 

Adds a reverse delay effect using the DelayLine worklet. It works by creating a delay line that is twice as long as the delay time and then reading the "mirror" index of the write node so the previously written nodes are read in reverse order. For example. if the delay line was length 10, this is what would be read at each write index:

|write index|read index|
|---|---|
|5|4|
|6|3|
|7|2|
|8|1|
|9|0|

I used this as a reference: https://ccrma.stanford.edu/~jingjiez/portfolio/echoing-harmonics/pdfs/A%20Pitch%20Shifting%20Reverse%20Echo%20Audio%20Effect.pdf

There is an example page at examples/reverseDelay.html

Let me know if you have any questions or if there is anything I can improve.


tambien:
awesome work! thanks for contributing, i will do a deeper review when I have a moment!

tambien:
super cool effect! 🙏🏻

PR #1399, opened 2026-01-19
feat: added width argument for AutoPanner

lukeklotz:
I wrote a couple tests for this, which passed. I also wrote a small toy project using this feature, which worked for me.


tambien:
Works great! thank you!

tambien:
Sorry actually realized that this is entirely redundant with the "depth" property which is inherited from the LFOEffect class and has the added benefit of being audio-rate instead of just a number. I'm going to revert this PR to avoid redundancy.

PR #1421, opened 2026-03-05
fix: use native ConstantSourceNode in CrossFade and StereoWidener

### LINKED ISSUE #1392: Avoid autoplay warnings from standardized-audio-context probes when constructing Effect/StereoEffect nodes on a suspended context
## Description

 - When instantiating Phaser, Reverb, Distortion (all Effect/StereoEffect subclasses) before any user gesture, Chrome logs “The AudioContext was not allowed to start…” repeatedly.
 - The warnings come from standardized-audio-context feature tests invoked by the wet/dry crossfade (getConstant() → testAudioBufferSourceNodeStartMethod…), which create an AudioBufferSourceNode and call start() while the shared context is suspended.
 - Other Tone nodes like Filter, Gain, Delay, Compressor don’t trigger these probes and remain quiet.

## Repro

```javascript
import * as Tone from "tone";
console.log("state before:", Tone.getContext().state); // "suspended"
const phaser = new Tone.Phaser(); // Chrome logs autoplay warning(s)
```

Run this in Chrome before any Tone.start()/user gesture; you’ll see the autoplay warnings. After a user gesture, the warnings stop.

## Request:

Please guard or defer these feature tests until the context is running (e.g., skip when context.state !== "running" or run after Tone.start()), or otherwise avoid triggering start() on a suspended context during construction of Effect/StereoEffect nodes to keep production consoles clean.

I noticed updating Tone.js to next (~15.3) from latest (~15.1) silenced the majority of the graph construction warning logs (because the internal implementation of ToneConstantSource added a guard for resuming suspended context), but some node warnings remain, and if the previous changes were intended to clean up those warnings, this should cover the remaining ones.

mxfng:
## Summary

`CrossFade` and `StereoWidener` each create a `ToneConstantSource` that outputs a static value of `1` to drive gain computations. After #1407, `ToneConstantSource` defers native node creation via `_onContextRunning`. When these classes are constructed on a suspended context, the constant source may not produce output, leaving both gain coefficients at `0`, silencing all audio through the node.

Every `Effect` subclass depends on `CrossFade` for dry/wet control, so this affects all effects in the library.

This PR replaces `ToneConstantSource` with a native `ConstantSourceNode` in both classes. Neither uses scheduling, start/stop lifecycle, or unit conversion. They just need a constant `1`, immediately, for the lifetime of the node.

Fixes #1392 (regression from #1407)

## Changes

- **CrossFade**: replace `ToneConstantSource` with native `ConstantSourceNode`
- **StereoWidener**: same fix, using `connect()` from `ToneAudioNode` to handle the native-to-Tone node connection

## Test plan

- [x] Component tests pass (487 passed, 0 failed)
- [x] Effect tests pass (325 passed, 0 failed)
- [x] TypeScript compiles cleanly via `ts:build`
- [ ] Verify no autoplay warnings in Chrome (original #1392 fix preserved)
- [ ] Verify effects produce audio when constructed on a suspended context

### DIFF
--- a/Tone/component/channel/CrossFade.ts
+++ b/Tone/component/channel/CrossFade.ts
@@ -9,7 +9,6 @@ import { optionsFromArguments } from "../../core/util/Defaults.js";
 import { readOnly } from "../../core/util/Interface.js";
 import { GainToAudio } from "../../signal/GainToAudio.js";
 import { Signal } from "../../signal/Signal.js";
-import { ToneConstantSource } from "../../signal/ToneConstantSource.js";
 
 interface CrossFadeOptions extends ToneAudioNodeOptions {
 	fade: NormalRange;
@@ -63,7 +62,7 @@ export class CrossFade extends ToneAudioNode<CrossFadeOptions> {
 	/**
 	 * The constant source which is used to control the panner
 	 */
-	private _constant: ToneConstantSource;
+	private _constant: ConstantSourceNode;
 
 	/**
 	 * The input which is at full level when fade = 0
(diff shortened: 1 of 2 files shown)


mxfng:
After digging deeper, I think the deferral path in `_contextStarted` does handle this correctly. The issue I was seeing is likely browser-level context suspension rather than the constant source lifecycle. Closing for now; will reopen with a minimal repro if I can isolate a specific failure in the deferral path.

PR #1453, opened 2026-06-30
fix(test): remove focused test modifiers

HURRAEY:
## Summary

- Remove focused Mocha modifiers from Gate and ToneConstantSource tests so those files run their full test coverage.
- Update the Gate above-threshold assertion to check for a positive output sample instead of requiring every sample to be positive, since the gate smoothing can leave the initial output at zero.

## Test plan

- [x] `npm run lint` in an LF checkout
- [x] `npx tsc --noEmit`
- [x] `npx eslint Tone/component/dynamics/Gate.test.ts Tone/signal/ToneConstantSource.test.ts`
- [x] `npx web-test-runner --config=./test/web-test-runner.config.js --coverage --group component`
- [x] `npx web-test-runner --config=./test/web-test-runner.config.js --coverage --group signal`

Note: I also ran the full browser suite locally on Windows with `npx web-test-runner --config=./test/web-test-runner.config.js --coverage`. The changed `component` and `signal` areas passed, but the full run reported unrelated failures in `Context`, `Oscillator`, `PulseOscillator`, and `PWMOscillator` tests.

### DIFF
--- a/Tone/component/dynamics/Gate.test.ts
+++ b/Tone/component/dynamics/Gate.test.ts
@@ -10,7 +10,7 @@ import { Gate } from "./Gate.js";
 describe("Gate", () => {
 	BasicTests(Gate);
 
-	it.only("matches a file", () => {
+	it("matches a file", () => {
 		return CompareToFile(
 			() => {
 				const gate = new Gate(-10, 0.1).toDestination();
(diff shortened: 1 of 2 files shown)


tambien:
thanks @HURRAEY

